PR #35, opened 2022-05-17
[fix] add mount umount lock to fix crash problem

fuxiuyin:
**What type of PR is this?**

/kind bug

OS sometime will crash when doing umount parallel to Lustre Volume. So, add a lock to serialize the mount/umount option.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


joe-atzinger (approved):
This is temporary until we can successfully get the core dumps and root cause.

PR #41, opened 2022-05-24
[feat] add mount options support

fuxiuyin:
**What type of PR is this?**

/kind feature

add mount options support and unit tests.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azurelustre/nodeserver.go
```diff
@@ -81,6 +82,12 @@ func (d *Driver) NodePublishVolume(
 	if req.GetReadonly() {
 		mountOptions = append(mountOptions, "ro")
 	}
+	for _, userMountFlag := range userMountFlags {
```
joe-atzinger, line 85: I think if we do the for-loop before line 82, we don't need to check inside the loop for the read-only option.  We can check once for existence before we try and append on line 82.

pkg/azurelustre/nodeserver_test.go
```diff
@@ -170,6 +170,42 @@ func TestNodePublishVolume(t *testing.T) {
...
 		},
+		{
+			desc: "Valid mount options",
+			req: csi.NodePublishVolumeRequest{VolumeCapability: &csi.VolumeCapability{AccessMode: &volumeCap, AccessType: &csi.VolumeCapability_Mount{
+				Mount: &csi.VolumeCapability_MountVolume{MountFlags: []string{"noatime", "noflock"}},
+			}},
+				VolumeId:          "vol_1",
+				TargetPath:        targetTest,
+				StagingTargetPath: sourceTest,
+				VolumeContext:     map[string]string{"mds-ip-address": "1.1.1.1", "fs-name": "lustrefs"},
```
joe-atzinger, line 181: "mgs-ip-address"

pkg/azurelustre/nodeserver_test.go
```diff
@@ -170,6 +170,42 @@ func TestNodePublishVolume(t *testing.T) {
...
+				Readonly:          false},
+			expectedErr: nil,
+		},
+		{
+			desc: "Valid mount options with readonly",
+			req: csi.NodePublishVolumeRequest{VolumeCapability: &csi.VolumeCapability{AccessMode: &volumeCap, AccessType: &csi.VolumeCapability_Mount{
+				Mount: &csi.VolumeCapability_MountVolume{MountFlags: []string{"noatime", "noflock"}},
```
joe-atzinger, line 188: The mount param is "flock", not "noflock".
Multiple places.

pkg/azurelustre/nodeserver_test.go
vinli-cn: Linux test failed because of "mds", pls fix it and check the test

pkg/azurelustre/nodeserver.go
fuxiuyin: Sorry, I'm not quite get your point. If we do the for-loop before check req.GetReadonly() and don't check the "ro" option. We still need to check if "ro" exists in the mountOptions when we check req.GetReadonly(). Otherwise, we may add two "ro" to the mountOptions.

joe-atzinger: You still need to check before adding req.ReadOnly.  If you move the for-loop to before that check, the loop no longer needs to check every time.

joe-atzinger: I have just investigated and found that go does not have "contains" functionality built-in, and you would need to iterate over the entire array again to check.  The current approach is straightforward enough to keep.  You can resolve this thread with no changes.

joe-atzinger (approved):

Issue #96, opened 2023-02-03
Health port needs to be configurable


joe-atzinger:
**Is your feature request related to a problem?/Why is this needed**
We encountered this with an internal user:
kube-system         csi-azurelustre-node-2ldpd                           0/3     Pending   0             76m
kube-system         csi-azurelustre-node-b2kmb                           0/3     Pending   0             76m
kube-system         csi-azurelustre-node-lfhmc                           0/3     Pending   0             76m

Events:
  Type     Reason            Age                 From               Message
  ----     ------            ----                ----               -------
  Warning  FailedScheduling  79m                 default-scheduler  0/3 nodes are available: 1 node(s) didn't have free ports for the requested pod ports. preemption: 0/3 nodes are available: 2 node(s) didn't match Pod's node affinity/selector, 3 node(s) didn't have free ports for the requested pod ports.
  Warning  FailedScheduling  24m (x10 over 74m)  default-scheduler  0/3 nodes are available: 1 node(s) didn't have free ports for the requested pod ports. preemption: 0/3 nodes are available: 2 node(s) didn't match Pod's node affinity/selector, 3 node(s) didn't have free ports for the requested pod ports.

**Describe the solution you'd like in detail**
We already parameterize other parts of the yaml spec.
We should also parameterize `containerPort: 29633` for the azurelustre container and liveness-probe.

**Describe alternatives you've considered**
We're going to run into other pods/containers running on specific ports.  We need to be able to work around them.

joe-atzinger:
Additionally, we need to pick a new default port.  csi-blob-node is using port 29633.

vinli-cn:
Thanks Joe! below is the reserved ports for Lustre CSI driver

controller
health: 29762
metrics: 29764

node
health: 29763
metrics: 29765

PR #97, opened 2023-02-03
Update CSI driver ports

vinli-cn:
**What type of PR is this?**
bug

Fixes #
#96 
**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/deploy/csi-azurelustre-controller.yaml
+++ b/deploy/csi-azurelustre-controller.yaml
@@ -74,10 +74,10 @@ spec:
             - containerPort: 29762
               name: healthz
               protocol: TCP
-            # Reserved port for metrics
-            # - containerPort: 29764
-            #   name: metrics
-            #   protocol: TCP
+          # Reserved port for metrics
+          # - containerPort: 29764
+          #   name: metrics
+          #   protocol: TCP
           livenessProbe:
             failureThreshold: 5
             httpGet:
--- a/deploy/csi-azurelustre-controller.yaml
+++ b/deploy/csi-azurelustre-controller.yaml
@@ -74,9 +74,10 @@ spec:
             - containerPort: 29762
               name: healthz
               protocol: TCP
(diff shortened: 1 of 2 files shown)


deploy/csi-azurelustre-controller.yaml
```diff
@@ -71,10 +71,10 @@ spec:
             - "-v=5"
             - "--endpoint=$(CSI_ENDPOINT)"
           ports:
-            - containerPort: 29632
+            - containerPort: 29762
               name: healthz
               protocol: TCP
-            - containerPort: 29634
+            - containerPort: 29764
```
fuxiuyin, line 77: The metrics port can be deleted. The metrics handler has been removed before in thie [PR 37](https://github.com/kubernetes-sigs/azurelustre-csi-driver/pull/37).

vinli-cn: Sure, I comment it out. Leave the port in deployment yaml so we won't forget the port num in the future

deploy/csi-azurelustre-node.yaml
```diff
@@ -95,9 +95,13 @@ spec:
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--nodeid=$(KUBE_NODE_NAME)"
           ports:
-            - containerPort: 29633
+            - containerPort: 29763
               name: healthz
               protocol: TCP
+          # Reserved port for metrics
```
fuxiuyin, line 101: Node shouldn't export this port. In blob controller is in aks internel and they can request the port to collect the metrics directly, so the controller exports the port. Node is in user's pool the metrics should be collected by user manually. For security reason I think. The port should be proxy by kubelet manually by `kubectl port-forward`.

fuxiuyin (approved):

PR #99, opened 2023-02-24
[bug] fix lnet

fuxiuyin:
**What type of PR is this?**
/kind bug

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurelustreplugin/entrypoint.sh
+++ b/pkg/azurelustreplugin/entrypoint.sh
@@ -94,7 +94,7 @@ if [[ "${installClientPackages}" == "yes" ]]; then
     modprobe -v lnet
     lnetctl lnet configure
 
-    echo "$(date -u) Determaing the default network interface."
+    echo "$(date -u) Determining the default network interface."
     # perl will be installed as dependency by luster client
     echo "$(date -u) Route table is:"
     ip route list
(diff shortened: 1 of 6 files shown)


pkg/azurelustreplugin/entrypoint.sh
```diff
@@ -78,18 +80,55 @@ if [[ "${installClientPackages}" == "yes" ]]; then
...
+
+    echo "$(date -u) Determaing the default network interface."
+    # perl will be installed as dependency by luster client
+    echo "$(date -u) Route table is:"
+    ip route list
+    default_interface=$(ip route list | perl -n -e'/default via [0-9.]+ dev ([0-9a-zA-Z]+) / && print $1')
+    echo "$(date -u) Default network interface is ${default_interface}"
+
+    if [[ "${default_interface}" == "" ]]; then
+      echo "$(date -u) Cannot determain the default network interface"
```
vinli-cn, line 105: determine?

pkg/azurelustreplugin/entrypoint.sh
```diff
@@ -78,18 +80,55 @@ if [[ "${installClientPackages}" == "yes" ]]; then
...
+      exit 1
+    fi
+
+    lnetctl net add --net tcp --if "${default_interface}"
```
vinli-cn, line 109: The default interface is fixed in bash script, any network changes could change the default interface?

fuxiuyin: If user wants to add new azure NIC. They needs to restart the VM. Then this script will be re-run when CSI is redeployed to this VM. User can change the default NIC but all azure NICs are in the same vnet and can access peered Lustre unless they have some weird network policy. If that's the case, there's no way to decide here it can be decided when mounting. But that should be an unlikely scenario to encounter.

PR #130, opened 2023-05-30
Use new SHA with LNET fix

### LINKED ISSUE #115: Remove workaround for LNET fix
**What happened**:
We applied workaround to fix LNET issue on Kubenet networking, we need to remove the workaround when the LNET issue is fixed in Lustre code

**What you expected to happen**:

**How to reproduce it**:

**Anything else we need to know?**:

**Environment**:
- CSI Driver version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

dabradley:
**What type of PR is this?**
/kind bug

The changes to allow LNET to work with correctly with namespaces have been completed. We no longer need the logic that was previously working around this issue. The changes here will ensure that all available ethernet interfaces are added during setup, which should prevent another issue where multiple nics were preventing the setup from adding the correct interface for AMLFS connectivity.

Fixes #115 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azurelustreplugin/entrypoint.sh
```diff
@@ -80,63 +118,42 @@ if [[ "${installClientPackages}" == "yes" ]]; then
...
-
   init_lnet="true"
-  
+
   if lsmod | grep "^lnet"; then
     if lnetctl net show --net tcp | grep interfaces; then
-      echo "$(date -u) LNet is loaded skip the load."
+      echo "$(date -u) LNet is loaded skip the load"
+      echo "$(date -u) Adding missing interfaces"
+      add_net_interfaces
```
fuxiuyin, line 127: LNet module is loaded in L138. So, the `lnetctl` command in `add_net_interfaces` will be executed before LNet is loaded. And will encounter some error I think.

dabradley: There are two logical flows here, one if the pod is being run on a node with lnet/lnetctl already loaded, or not.

The if statement here means this will run if lnet/lnetctl is already loaded (L123-124). This just makes sure that all available ethernet nics are added.

Line 138, which loads lnet, is inside a conditional to only run if lnet hasn't already been loaded, specifically when init_lnet is true (this block makes sure that's skipped on L128).

fuxiuyin (approved):

PR #138, opened 2023-10-15
Upgrade driver dependencies and images

dabradley:
We have been getting flagged for CVEs and other issues due to using older versions in this CSI driver. This commit upgrades all of the affected driver dependencies, as well as the sidecar images and the driver base image.

Some minor code changes were necessary to migrate from the deprecated protobuf library to the currently supported version.

Replace rules were added to ensure the correct kubernetes library versions are used, as the library has a non-standard dependency configuration which would otherwise cause the go commands to always attempt to downgrade the library.

It is now possible to run some of the tests that were being skipped, as the CSI driver now supports that functionality. The remaining tests that are being skipped are due to the driver's lack of proper dynamic provisioning functionality.

There are still some tests that were never in the skipped list that fail sporadically due to concurrency issues, because of the same dynamic provisioning concerns. These will be left as is until proper dynamic provisioning functional is added, as it is already a high priority roadmap item.

**What type of PR is this?**
/kind bug

We have been getting flagged for CVEs and other issues due to using older versions in this CSI driver. This commit upgrades all of the affected driver dependencies, as well as the sidecar images and the driver base image.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/lgtm

PR #139, opened 2023-10-17
Allow CSI driver to function entirely within container

dabradley:
**What type of PR is this?**
/kind bug

We have been hitting numerous customer issues due to the state
that is left behind on a node when we install the CSI driver
using the current process, especially due to apt having issues
with differing versions of the driver.

This change removes the need for the driver to be installed
directly into the node's filesystem. Since the lustre driver
is installed into the kernel as a module, it is sufficient
for our driver container to have the appropriate permissions
and mount points solely for loading the module and operating
as a CSI driver.

This driver is configured to unload the kernel module when
the pod is deleted, allowing for a seamless upgrade experience
in future versions, where the kernel module is loaded from
a clean system, and without leaving permanent installation
artifacts around in the node's root filesystem.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


deploy/csi-azurelustre-node.yaml
```diff
@@ -157,55 +142,23 @@ spec:
...
-          name: host-run-udev
-        - hostPath:
-            path: /lib
-            type: Directory
-          name: host-lib
-        - hostPath:
-            path: /lib64
-            type: Directory
-          name: host-lib64
-        - hostPath:
```
joe-atzinger, line 204: I believe the upcall for configuration changes from the kernel to userspace requires the binaries to be accessible in the host filesystem.  This manifests as a negative 10x perf impact in IOPS tests when it doesn't work.

dabradley: Note to myself: The values in question are the following:
```bash
# lctl get_param -R *.*.max_rpcs_in_flight
mdc.lustrefs-MDT0000-mdc-ffff88e8c16cc800.max_rpcs_in_flight=128
osc.lustrefs-OST0000-osc-ffff88e8c16cc800.max_rpcs_in_flight=32
```
By default they are `8`. If we want to do this approach, we need to make sure the driver is correctly configuring and using those values. Also check on any other non-default values that were configured on the nodes to ensure they're being applied correctly as well.

Issue #141, opened 2023-10-27
CrashLoopBackOff  in some csi-azurelustre-node pods


varameshgithub:
**What happened**:

We are facing intermittent CrashLoopBackOff in some of the csi-azurelustre-node* pods in AKS worker nodes. Not facing this issue very often but occurs intermittently usually when we scale up the number of nodes to a higher number (say from 4 to 300 nodes) and sometimes during onboarding a new AKS cluster. While creating the new node some of the  csi-azurelustre-node* goes into CrashLoopBackOff and restarting multiple times with the same error.

_2023-10-22T07:30:46.79512136Z stderr F + [[ yes == \y\e\s ]]
2023-10-22T07:30:46.795371005Z stderr F ++ uname -r
2023-10-22T07:30:46.795924737Z stderr F + kernelVersion=5.15.0-1049-azure
2023-10-22T07:30:46.796250124Z stderr F ++ date -u
2023-10-22T07:30:46.796921159Z stdout F Sun Oct 22 07:30:46 UTC 2023 Installing Lustre client packages for OS=jammy, kernel=5.15.0-1049-azure
2023-10-22T07:30:46.796930232Z stderr F + echo 'Sun Oct 22 07:30:46 UTC 2023 Installing Lustre client packages for OS=jammy, kernel=5.15.0-1049-azure '
2023-10-22T07:30:46.796940876Z stderr F + '[' '!' -f /etc/apt/sources.list.d/amlfs.list ']'
2023-10-22T07:30:46.79728655Z stderr F ++ date -u
2023-10-22T07:30:46.798187775Z stdout F Sun Oct 22 07:30:46 UTC 2023 Installing Lustre client modules: amlfs-lustre-client-2.15.1-29-gbae0abe=5.15.0-1049-azure
2023-10-22T07:30:46.798192133Z stderr F + echo 'Sun Oct 22 07:30:46 UTC 2023 Installing Lustre client modules: amlfs-lustre-client-2.15.1-29-gbae0abe=5.15.0-1049-azure'
2023-10-22T07:30:46.798198395Z stderr F + DEBIAN_FRONTEND=noninteractive
2023-10-22T07:30:46.798202182Z stderr F + apt install -y --no-install-recommends -o DPkg::options::=--force-confdef -o DPkg::options::=--force-confold amlfs-lustre-client-2.15.1-29-gbae0abe=5.15.0-1049-azure
[...]

PR #142, opened 2023-10-27
Changes to entrypoint for better resilience

varameshgithub:
**What type of PR is this?**
Bug

In addition to check /etc/apt/sources.list.d/amlfs.list exists we need to check if the apt-update is successful. This will help us recover the pod in-case of restarts during apt-update or failure in apt-update there by not leading the pod in an irrecoverable state.

Issue  https://github.com/kubernetes-sigs/azurelustre-csi-driver/issues/141


joe-atzinger:
Just waiting for testing to complete before I approve and merge.

joe-atzinger (approved):

PR #149, opened 2023-11-20
Uninstall existing driver if upgrade fails

dabradley:
We currently don't have a supported in-place upgrade path for users because of the way our packages are configured. There are a number of ways that the installed packages can get into bad states that apt (and even dpkg directly) refuses to work around without manual intervention.

To avoid this behavior, this commit adds limited retries to the package installation step that, in the instance of a failed client install, will attempt to unload any running Lustre modules, remove the installed packages, and retry the installation.

In addition, when the pod is terminated, it will attempt to unload the Lustre modules, but will not fail if it cannot unload them (such as when there are existing pods running which have active Lustre mounts).

If it cannot unload the existing modules, it will log a warning that the new version will likely not be running, but will not fail. Additionally, if it cannot install the desired packages, it will log a warning that it could not install correctly, but will continue to attempt to start the driver.

**What type of PR is this?**
/kind bug

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

Tested this with:
- upgrading and downgrading the SHA (succeeds after first failed try)
- with a known bad state (succeeds after first failed try)
- with an existing, correctly installed driver of the expected version (succeeds on first try with no retries)
- from a clean state with no driver installed (succeeds on first try with no retries)

**Release note**:
```
none
```

### DIFF
--- a/deploy/csi-azurelustre-node.yaml
+++ b/deploy/csi-azurelustre-node.yaml
@@ -58,10 +58,6 @@ spec:
             - --csi-address=$(ADDRESS)
             - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
             - --v=2
-          lifecycle:
-            preStop:
-              exec:
-                command: ["/bin/sh", "-c", "rm -rf /registration/azurelustre.csi.azure.com-reg.sock /csi/csi.sock"]
           livenessProbe:
             exec:
               command:
@@ -90,6 +86,10 @@ spec:
         - name: azurelustre
           image: mcr.microsoft.com/oss/kubernetes-csi/azurelustre-csi:v0.1.12
           imagePullPolicy: IfNotPresent
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "lustre_rmmod"]
           args:
             - "-v=5"
             - "--endpoint=$(CSI_ENDPOINT)"
@@ -117,6 +117,8 @@ spec:
               value: "yes"
             - name: LUSTRE_VERSION
               value: "2.15.1"
+            - name: CLIENT_SHA_SUFFIX
+              value: "33-g0168b83"
             - name: KUBE_NODE_NAME
               valueFrom:
                 fieldRef:
--- a/pkg/azurelustreplugin/entrypoint.sh
+++ b/pkg/azurelustreplugin/entrypoint.sh
@@ -115,19 +115,27 @@ if [[ "${installClientPackages}" == "yes" ]]; then
   echo "$(date -u) Installing Lustre client modules: ${pkgName}=${kernelVersion}"
 
   tries=3
+  sleep_before_retry=15
   install_success=false
   while [[ tries -gt 0 ]]; do
     # grub issue
     # https://stackoverflow.com/questions/40748363/virtual-machine-apt-get-grub-issue/40751712
-    if ! DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends -o DPkg::options::="--force-confdef" -o DPkg::options::="--force-confold" \
+    if ! DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends -o DPkg::options::="--force-confdef" -o DPkg::options::="--force-confold" \
       ${pkgName}=${kernelVersion}; then
       echo "$(date -u) Error installing Lustre client modules. Will try removing existing versions"
-      if ! lustre_rmmod; then
+      # Check if lustre_rmmod is available, attempt to unload the modules if so.
+      # If modules are already uninstalled, this will still pass
+      if type lustre_rmmod >/dev/null 2>&1 && ! lustre_rmmod; then
         echo "$(date -u) Error: Unable to unload running module. Are there still mounted Lustre filesystems on this node? Old Lustre client version may continue running."
       fi
+      if existing_versions=$(dpkg-query --showformat=' ${Package}=${Version}' --show '*lustre-client*'); then
+        echo  "$(date -u) The following existing versions of the Lustre client are installed and will be removed:${existing_versions}"
+      fi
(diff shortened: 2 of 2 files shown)


dabradley:
@joe-atzinger I'd appreciate if you could get a chance to look at this one. This would be a nice quality of life improvement since customers seem to get stuck with bad upgrade states a lot

pkg/azurelustreplugin/entrypoint.sh
```diff
@@ -108,15 +111,36 @@ if [[ "${installClientPackages}" == "yes" ]]; then
...
-    ${pkgName}=${kernelVersion}
+  tries=3
+  install_success=false
+  while [[ tries -gt 0 ]]; do
+    # grub issue
+    # https://stackoverflow.com/questions/40748363/virtual-machine-apt-get-grub-issue/40751712
+    if ! DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends -o DPkg::options::="--force-confdef" -o DPkg::options::="--force-confold" \
+      ${pkgName}=${kernelVersion}; then
+      echo "$(date -u) Error installing Lustre client modules. Will try removing existing versions"
+      if ! lustre_rmmod; then
```
joe-atzinger, line 125: Will this fail if no modules / packages are installed?
Thinking of a case where install failed for some reason, but it wasn't because modules were already installed - this would be a misleading error message.

pkg/azurelustreplugin/entrypoint.sh
```diff
@@ -108,15 +111,36 @@ if [[ "${installClientPackages}" == "yes" ]]; then
...
-  # grub issue
-  # https://stackoverflow.com/questions/40748363/virtual-machine-apt-get-grub-issue/40751712
-  DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends -o DPkg::options::="--force-confdef" -o DPkg::options::="--force-confold" \
-    ${pkgName}=${kernelVersion}
+  tries=3
+  install_success=false
+  while [[ tries -gt 0 ]]; do
+    # grub issue
+    # https://stackoverflow.com/questions/40748363/virtual-machine-apt-get-grub-issue/40751712
+    if ! DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends -o DPkg::options::="--force-confdef" -o DPkg::options::="--force-confold" \
```
joe-atzinger, line 122: I think you want apt-get, right?  To avoid the warning about an unstable apt API.

pkg/azurelustreplugin/entrypoint.sh
joe-atzinger: If you had a separate command to check for any installed lustre client packages first, you could then print out the pkg version with a helpful bread crumb.

pkg/azurelustreplugin/entrypoint.sh
```diff
@@ -108,15 +111,36 @@ if [[ "${installClientPackages}" == "yes" ]]; then
...
+    fi
+  done
 
-  echo "$(date -u) Installed Lustre client packages."
+  echo "$(date -u) Install success: ${install_success}, Tries left: ${tries}"
+
+  if ! ${install_success}; then
+    echo "$(date -u) Error: Could not install necessary Lustre drivers!"
+  else
+    echo "$(date -u) Installed Lustre client packages."
```
joe-atzinger, line 142: Probably worth putting the pkg version in the echo for sanity's sake.

deploy/csi-azurelustre-node.yaml
```diff
@@ -58,10 +58,6 @@ spec:
             - --csi-address=$(ADDRESS)
             - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
             - --v=2
-          lifecycle:
-            preStop:
-              exec:
-                command: ["/bin/sh", "-c", "rm -rf /registration/azurelustre.csi.azure.com-reg.sock /csi/csi.sock"]
```
joe-atzinger, line 64: Does this no longer need to be removed?

joe-atzinger (changes requested):
Done reviewing first pass.

pkg/azurelustreplugin/entrypoint.sh
```diff
@@ -108,15 +111,36 @@ if [[ "${installClientPackages}" == "yes" ]]; then
...
+      if ! lustre_rmmod; then
+        echo "$(date -u) Error: Unable to unload running module. Are there still mounted Lustre filesystems on this node? Old Lustre client version may continue running."
+      fi
+      echo "$(date -u) Uninstalling existing Lustre client versions."
+      apt remove --purge -y '*lustre-client*' || true
+      tries=$((tries - 1))
+    else
+      install_success=true
+      break
+    fi
```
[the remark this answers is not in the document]
dabradley: Fair enough, I'll add that

pkg/azurelustreplugin/entrypoint.sh
dabradley: This line doesn't change at all in this PR (just moved into the retry block. But I'll try changing it over because that message is annoying

pkg/azurelustreplugin/entrypoint.sh
dabradley: I'll see about having it react better if the modules aren't installed/loaded.

As far as checking for installed packages first, an earlier attempt at this did that (to be able to check if we need to upgrade or not). In a lot of cases, we get back a mishmash of mixed up versions (and the general package without a specified version). If it's useful, I could have it at least print out what apt lists as the installed versions just to give someone debugging a better idea of what's going on.

joe-atzinger: I think you'll benefit from more breadcrumbs if you can get them.

joe-atzinger (approved):

PR #151, opened 2023-11-27
Add sub-dir functionality

dabradley:
This commit adds the ability for pods to bind to a Lustre volume in a specified subdirectory. The pod that binds the volume in this way can only see that path and the paths under it. It has no access or visibility to other paths above it in the Lustre filesystem.

This has been implemented similarly to other CSI drivers to meet user expectations, but has the added benefit of allowing the user to leverage metadata from both the pvc/pv metadata and the pod's metadata as well.

Numerous checks have been put in place to ensure that data outside of the requested subdirectory is not deleted and that read only volumes are handled reasonably.

The long-haul tests have had compatible subdirectory values set to exercise this functionality while still maintaining its expectations as to data persistence, etc.

This is largely based on the implementations in the nfs and smb drivers, but it operates on a per-pod basis.

Design document (for MSFT employees) here: https://dev.azure.com/msazure/One/_git/Avere-laaso-doc/pullrequest/9149798

**What type of PR is this?**
/kind feature

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/docs/examples/pv_subdir.yaml
+++ b/docs/examples/pv_subdir.yaml
@@ -1,4 +1,7 @@
 ---
+# The subdirectory functionality is not yet released; these files are
+# provided as examples for the next version of this driver. This message
+# will be removed at release time when we make this new release the default
 apiVersion: v1
 kind: PersistentVolume
 metadata:
@@ -21,6 +24,7 @@ spec:
       # The subdirectory to create and mount per pod (cannot use pvc / pv metadata in static volumes)
       sub-dir: "${pod.metadata.namespace}_${pod.metadata.name}"
       # Whether the created subdirectory should be retained or deleted at pod exit
+      # This parameter is required if "sub-dir" is provided
       retain-sub-dir: "false"
     # Make sure this VolumeID is unique in the cluster
     volumeHandle: ${UNIQUE_IDENTIFIER_VOLUME_ID}
--- a/docs/examples/pv_subdir.yaml
+++ b/docs/examples/pv_subdir.yaml
@@ -1,0 +1,32 @@
+---
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  # The name of the PV
+  name: pv-lustre-subdir
+spec:
+  accessModes:
+    - ReadWriteMany
+  capacity:
+    # This field should be the true size of the Azure Lustre you want
+    # to used. So that, k8s can allocate resources better.
+    storage: 4Ti
+  csi:
+    driver: azurelustre.csi.azure.com
+    volumeAttributes:
+      # The file system name of the existing Lustre
+      fs-name: ${EXISTING_LUSTRE_FS_NAME}
+      # The IP address of the existing Lustre
+      mgs-ip-address: ${EXISTING_LUSTRE_IP_ADDRESS}
+      # The subdirectory to create and mount per pod (cannot use pvc / pv metadata in static volumes)
+      sub-dir: "${pod.metadata.namespace}_${pod.metadata.name}"
+      # Whether the created subdirectory should be retained or deleted at pod exit
+      retain-sub-dir: "false"
+    # Make sure this VolumeID is unique in the cluster
+    volumeHandle: ${UNIQUE_IDENTIFIER_VOLUME_ID}
+  # "Delete" is not supported in static provisioning PV
+  mountOptions:
+    - noatime
+    - flock
+  persistentVolumeReclaimPolicy: Retain
+  storageClassName: ""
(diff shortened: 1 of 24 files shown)


docs/examples/storageclass_existing_lustre_subdir.yaml
dabradley, line 1: Should these yaml files be held back until the image is released so that someone checking out the repo won't get them before the functionality is available?

dabradley:
@joe-atzinger This is the major feature I'm trying to get out the door. Let me know if you have any feedback on this so I can try to get this one moving

docs/examples/storageclass_existing_lustre_subdir.yaml
joe-atzinger: I don't think they will work as expected if applied, right?  You can add a comment indicating that the functionality isn't ready yet if you want.

docs/examples/pv_subdir.yaml
```diff
@@ -0,0 +1,32 @@
+---
```
joe-atzinger, line 1: If we're providing examples for a PV now (a good thing), then we should make sure to update the customer-facing docs with instructions for doing this.

pkg/azurelustre/nodeserver.go
```diff
@@ -357,3 +514,359 @@ func (d *Driver) ensureMountPoint(target string) (bool, error) {
...
+			subDir,
+			retainSubDir),
+	}
+
+	if volFromID != nil && *volFromID != *vol {
+		klog.Warningf("volume context does not match values in volume id for volume %q", volumeID)
+	}
+
+	return vol, nil
+}
```
joe-atzinger, line 872: We spoke offline re: this approach vs NodeStage/UnstageVolume.  It was good and useful information that should probably end up in here with a couple commented-out decls of NodeStage/UnstageVolume.  That way no one tries to do that later.

joe-atzinger (changes requested):
Mostly requesting some clarifying comments before approval.

test/long-haul/cleanup/cleanupjob.yaml
```diff
@@ -7,6 +7,8 @@ provisioner: azurelustre.csi.azure.com
 parameters:
   mgs-ip-address: "{lustre_fs_ip}"
   fs-name: "{lustre_fs_name}"
+  sub-dir: "longhaul"
```
dabradley, line 10: Is this a concern for a shared cluster with testing? Uniquify?

t-mialve: I know we've had setting up a cni longhaul pipeline on the backburner so maybe uniquify so we can set that up and point to the same cluster?

But we can also set up a 2nd cluster so leaving it as-is is fine too

pkg/azurelustre/nodeserver.go
```diff
@@ -357,3 +514,359 @@ func (d *Driver) ensureMountPoint(target string) (bool, error) {
...
+	}()
+
+	internalVolumePath, err := getInternalVolumePath(d.workingMountDir, vol, subDirPath)
+	if err != nil {
+		return err
+	}
+
+	klog.V(2).Infof("Making subdirectory at %q", internalVolumePath)
+
+	if err := os.MkdirAll(internalVolumePath, 0775); err != nil {
```
t-mialve, line 536: Will we want to allow for custom subdir perms down the road, or are we happy with read-only and this being the only options?

t-mialve (approved):

docs/examples/storageclass_existing_lustre_subdir.yaml
dabradley: Done

docs/examples/pv_subdir.yaml
dabradley: Noted for documentation

pkg/azurelustre/nodeserver.go
dabradley: Done

test/long-haul/cleanup/cleanupjob.yaml
dabradley: The existing logic runs this task to rm -rf everything on the volume, so if it anything this is now less destructive. We can revisit the general approach when extending this testing

pkg/azurelustre/nodeserver.go
dabradley: This is a good point, I'll note it for consideration as a separate change

joe-atzinger (approved):

PR #155, opened 2024-01-04
Correct per-publish/unpublish lock

dabradley:
Locking at the volume level was necessary for the original proof of concept that used stage/unstage steps with a global path. However, with the lock being used per-publish/unpublish step, it was causing unnecessary slowdowns when many pods were created or deleted for a given volume. This commit changes the behavior to lock only for a specific pod's directory, which will avoid this delay across individual pods for a single volume.
    

**What type of PR is this?**
/kind bug 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurelustre/nodeserver.go
+++ b/pkg/azurelustre/nodeserver.go
@@ -102,12 +102,13 @@ func (d *Driver) NodePublishVolume(
 		}
 	}
 
-	if acquired := d.volumeLocks.TryAcquire(vol.id); !acquired {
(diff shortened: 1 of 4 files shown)


andyzhangx:
refer to https://github.com/kubernetes-csi/csi-driver-nfs/pull/574

andyzhangx (approved):
/lgtm

PR #156, opened 2024-01-11
Skip subdirectory creation for invalid volume ID

dabradley:
The only reasonable way that the unpublish command can determine whether to keep or delete the subdirectory, if given, is to bake this information into the volume ID, as the original volume context is not passed to volume unpublish calls. For dynamically created volumes, this is always consistent. However, if the end user defines their own volumes, but does not follow the expected volume handle / volume ID form, that information is not available. In this instance, if the user sets the retain-sub-dir value on a manually created volume to 'false', we will error out and refuse to create the volume. If they request that the subdirectory be retained, we will treat the volume as we do a read-only volume, where we will attempt to mount the given subdirectory, but will not do any logic associated with creating or deleting that subdirectory. As with a read-only volume, if that subdirectory does not exist on the cluster, the mount will fail with a reasonable error.

**What type of PR is this?**
/kind bug

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

This PR is based on https://github.com/kubernetes-sigs/azurelustre-csi-driver/pull/155 and will be in draft until that is merged.

**Release note**:
```
none
```

### DIFF
--- a/docs/examples/pv_subdir.yaml
+++ b/docs/examples/pv_subdir.yaml
@@ -27,6 +27,11 @@ spec:
       # This parameter is required if "sub-dir" is provided
       retain-sub-dir: "false"
     # Make sure this VolumeID is unique in the cluster
+    # volumeHandle format: {volume-name}#{fs-name}#{mgs-ip-address}#{sub-dir}#{retain-sub-dir}
+    # Example: vol_1#lustrefs#0.0.0.0#example_sub_dir#true
+    # If the value does not match this format, the driver will not attempt to create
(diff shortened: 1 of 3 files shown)


dabradley:
@joe-atzinger Could you take a look at this PR? It's based on https://github.com/kubernetes-sigs/azurelustre-csi-driver/pull/155 so I'm leaving it a draft, but I'd like this to go in as soon as that's done

joe-atzinger (approved):
/lgtm

PR #159, opened 2024-02-06
Remove retain-sub-dir option

dabradley:
The retain-sub-dir feature is intended to allow the user to have their subdirectory data automatically cleaned up when a pod unmounts from the Lustre filesystem. However, based on test, in its current form it allows the user to easily get themselves into a bad state or to lose data. For this release, we will maintain the subdirectory creation and mounting logic, but will wait on any customer feedback before re-enabling this functionality based on expected use cases.

**What type of PR is this?**
/kind bug

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azurelustre/__debug_bin1101564440
csmuell, line 1: What is the use case for these debug files?

t-mialve (approved):

PR #185, opened 2024-11-08
Add skip for interfaces in unknown state to entrypoint script

t-mialve:
**What type of PR is this?**

/kind bug

Skips interfaces that are in an UNKNOWN state. If these aren't skipped it may cause extra client IPs to be added.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurelustreplugin/entrypoint.sh
+++ b/pkg/azurelustreplugin/entrypoint.sh
@@ -32,12 +32,14 @@ function add_net_interfaces() {
   ethernet_interfaces=()
   for interface in $interface_list; do
     interface_info=$(ip link show "${interface}")
+    # Skip interfaces that are not needed
     if [[ "$interface_info" =~ 'SLAVE' ]]; then
       echo "$(date -u) Not adding slave interface: ${interface}"
-      continue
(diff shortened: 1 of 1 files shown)


pkg/azurelustreplugin/entrypoint.sh
```diff
@@ -41,6 +41,9 @@ function add_net_interfaces() {
     elif [[ "$interface_info" =~ 'link/ether' ]]; then
       echo "$(date -u) Including ethernet interface: ${interface}"
       ethernet_interfaces+=("$interface")
+    elif [[ "$interface_info" =~ 'UNKNOWN' ]]; then
```
jeffbearer, line 44: Is there a list of all (common) kinds of interfaces that we should consult and also exclude?   Token Ring, ATM, Sneakernet?

pkg/azurelustreplugin/entrypoint.sh
```diff
@@ -41,6 +41,9 @@ function add_net_interfaces() {
...
+      echo "$(date -u) Not adding state unknown interface: ${interface}"
+      continue
```
t-mialve, line 46: Remove the continues

PR #196, opened 2025-01-06
Add fsGroupPolicy to driver

dabradley:
In order to be able to use non-root accounts, it is important for the driver to allow k8s to change group ids in the cluster. Adding this value enables this ability so that in all circumstances, a pod can run with the expected filesystem group ACLs set.

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


t-mialve (approved):
/lgtm

PR #256, opened 2026-01-15
Clean up update-gomod.sh script

dabradley:
**What type of PR is this?**
/kind cleanup

This script allows developers to upgrade the kubernetes dependencies (since their versioning is effectively broken and requires explicit 'replace' statements in the go.mod file). This PR fixes some linter issues and ensures that all necessary go mod commands are called.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/hack/update-gomod.sh
+++ b/hack/update-gomod.sh
@@ -18,19 +18,29 @@ set -euo pipefail
 
 VERSION=${1#"v"}
 if [ -z "$VERSION" ]; then
-    echo "Must specify version!"
+    echo "Please specify the Kubernetes version: e.g."
+    echo "./update-gomod.sh v1.32.11"
     exit 1
 fi
-MODS=($(
-    curl -sS https://raw.githubusercontent.com/kubernetes/kubernetes/v${VERSION}/go.mod |
+
+# Determine which imports must be replaced
+mapfile -t MODS < <(
+    curl -sS "https://raw.githubusercontent.com/kubernetes/kubernetes/v${VERSION}/go.mod" |
     sed -n 's|.*k8s.io/\(.*\) => ./staging/src/k8s.io/.*|k8s.io/\1|p'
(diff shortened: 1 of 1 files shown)


jeffbearer:
Nice cleanup — the quoting hardening, `mapfile` over array-from-subshell, curl error handling, and the added `go mod` steps at the end are all solid improvements.

One small issue worth catching while we're in here: the new friendly error message is unreachable for the most common failure mode (running the script with no arguments at all).

The line

```bash
VERSION=${1#"v"}
```

trips `set -u` *before* the `[[ -z "${VERSION}" ]]` check runs, because `$1` is unset when no arg is passed. Confirmed empirically:

```
$ ./update-gomod.sh
./update-gomod.sh: line 19: 1: unbound variable
```

So the user sees a bare bash error instead of the helpful "Please specify the Kubernetes version: e.g. ./update-gomod.sh v1.32.11" message that this PR added. The pre-existing `if [ -z "$VERSION" ]` had the same defect, but since this PR is specifically improving the message, it's worth making it actually reachable.

Two-line fix:

```bash
VERSION="${1:-}"
VERSION="${VERSION#v}"
```

`${1:-}` defaults `$1` to empty string if unset (defusing `set -u`); then the prefix strip runs on a guaranteed-set variable. After that, the existing empty-check catches both "no arg" and "empty arg" cases and prints the friendly message.

I verified the happy path is unchanged — ran both the PR script and the patched script against `v1.32.11` in a temp module, output and resulting `go.mod` are bit-for-bit identical.

dabradley:
@jeffbearer Thanks for catching that. I thought I'd tested that but obviously not the final version. Updated and pushed

jeffbearer (approved):
/lgtm
/approve

PR #257, opened 2026-01-15
Avoid errant warning in volume ID check

dabradley:
**What type of PR is this?**

The conditional in this case was always logging an issue even when there wasn't one. This didn't cause behavior issues but did make reading the log output confusing. The PR also makes sure the tests for this code are correct and complete, though it's not feasible to test the logging directly.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


jeffbearer (approved):
LGTM

jeffbearer (approved):
/lgtm
/approve

PR #272, opened 2026-04-07
Add Workload Identity support for AKS extension onboarding

jeffbearer:
**What type of PR is this?**
/kind feature

Adds Workload Identity support to the Azure Lustre CSI driver. This is a prerequisite for AKS extension onboarding per the [Partner Onboarding Security Requirements](https://msazure.visualstudio.com/One/_wiki/wikis/One.wiki/102775/Partner-Onboarding-Process), which require all extensions to support workload identity.

### Changes

**Credential handling** (`pkg/azurelustre/azurelustre.go`):
- Add `newAzureCredential()` that creates `WorkloadIdentityCredential` when `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, and `AZURE_FEDERATED_TOKEN_FILE` env vars are present (injected by the AKS workload identity webhook)
- Falls back to `DefaultAzureCredential` when workload identity is not configured, preserving backward compatibility
- Adds logging to indicate which credential type is being used

**Deploy manifests**:
- Add `azure.workload.identity/use: "true"` label to controller and node service accounts with commented annotation placeholders for `client-id`/`tenant-id`
- Add `azure.workload.identity/use: "true"` label to controller Deployment and node DaemonSet pod templates (the webhook matches on pod labels)
- Add pod-level `seccompProfile: RuntimeDefault` (PSS Baseline requirement)
- Add `capabilities.drop: [ALL]` to sidecar containers (csi-provisioner, liveness-probe, node-driver-registrar)
- Node `azurelustre` container retains `privileged: true` for kernel module loading and mount operations (same pattern as blob-csi-driver)

**Documentation**:
- New `docs/workload-identity.md` with prerequisites, step-by-step setup guide
- Updated `docs/driver-parameters.md` to reference workload identity docs

Fixes AB#35653648

**Requirements**:
- [x] includes documentation
- [x] adds unit tests

### Testing performed

**Local/CI:**
- `make unit-test`: ✅ 89.3% coverage, all passing (3 new tests for `newAzureCredential()`)
- `make sanity-test-local`: ✅ 30/30 passed
- `make azurelustre`: ✅ Binary builds

**Cluster validation** (MLT-10, canadacentral, custom image from branch):
- ✅ Static mount with PSS changes — Lustre mounted, data read/write works
- ✅ DefaultAzureCredential fallback — controller ran with managed identity when WI not annotated
- ✅ Workload identity injection — all 3 env vars + projected token volume injected by webhook
- ✅ Driver log confirmed: `using workload identity for authentication (tenantID: ..., clientID: ...)`
- ✅ **Dynamic provisioning via WI** — AMLFS cluster (4 TiB) created through ARM API calls authenticated via federated service account token. PVC bound successfully.

### Design decisions
- Workload identity is **optional** — falls back to `DefaultAzureCredential` (managed identity from `/etc/kubernetes/azure.json`) when not configured
- Azure auth is only needed for **dynamic provisioning** (ARM API calls). Static provisioning uses kernel Lustre mounts and does not require Azure authentication
- The `/etc/kubernetes/azure.json` hostPath mount is still needed for subscription ID, resource group, location context
- Follows patterns from `kubernetes-sigs/blob-csi-driver`

**Release note**:
```
feat: Add Workload Identity support for Azure authentication, enabling federated service account token-based auth as an alternative to managed identity. Required for AKS extension onboarding.
```

### DIFF
--- a/deploy/csi-azurelustre-controller.yaml
+++ b/deploy/csi-azurelustre-controller.yaml
@@ -13,11 +13,15 @@ spec:
     metadata:
       labels:
         app: csi-azurelustre-controller
+        azure.workload.identity/use: "true"
     spec:
       serviceAccountName: csi-azurelustre-controller-sa
       nodeSelector:
         kubernetes.io/os: linux  # add "kubernetes.io/role: master" to run controller on master node
       priorityClassName: system-cluster-critical
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
       tolerations:
         - key: "node-role.kubernetes.io/master"
           operator: "Exists"
@@ -50,6 +54,10 @@ spec:
             requests:
               cpu: 10m
               memory: 20Mi
+          securityContext:
+            capabilities:
+              drop:
+                - ALL
         - name: liveness-probe
           image: mcr.microsoft.com/oss/kubernetes-csi/livenessprobe:v2.15.0
           args:
@@ -66,6 +74,10 @@ spec:
             requests:
               cpu: 10m
               memory: 20Mi
+          securityContext:
+            capabilities:
+              drop:
+                - ALL
         - name: azurelustre
           image: mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.4.0-jammy
           imagePullPolicy: Always
--- a/deploy/csi-azurelustre-node-jammy.yaml
+++ b/deploy/csi-azurelustre-node-jammy.yaml
@@ -18,6 +18,7 @@ spec:
       labels:
         app: csi-azurelustre-node
         flavor: jammy
+        azure.workload.identity/use: "true"
     spec:
       hostNetwork: true
       dnsPolicy: Default
@@ -39,6 +40,9 @@ spec:
                       - Ubuntu2204
                       - Ubuntu2004  # To support CVM nodes still on 20.04
       priorityClassName: system-node-critical
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
       tolerations:
         - operator: "Exists"
       containers:
@@ -59,6 +63,10 @@ spec:
             requests:
               cpu: 10m
               memory: 20Mi
+          securityContext:
+            capabilities:
+              drop:
+                - ALL
         - name: node-driver-registrar
           image: mcr.microsoft.com/oss/kubernetes-csi/csi-node-driver-registrar:v2.13.0
           args:
@@ -90,6 +98,10 @@ spec:
             requests:
               cpu: 10m
               memory: 20Mi
+          securityContext:
+            capabilities:
+              drop:
+                - ALL
         - name: azurelustre
           image: mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.4.0-jammy
           imagePullPolicy: Always
--- a/deploy/csi-azurelustre-node-noble.yaml
+++ b/deploy/csi-azurelustre-node-noble.yaml
@@ -18,6 +18,7 @@ spec:
       labels:
         app: csi-azurelustre-node
         flavor: noble
+        azure.workload.identity/use: "true"
     spec:
       hostNetwork: true
       dnsPolicy: Default
@@ -35,6 +36,9 @@ spec:
                     values:
                       - virtual-kubelet
       priorityClassName: system-node-critical
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
       tolerations:
         - operator: "Exists"
       containers:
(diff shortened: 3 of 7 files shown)


jeffbearer:
## Test Report: Workload Identity Manifest Changes

Validated on MLT-10 (canadacentral) using stock MCR image v0.4.0 — zero Go code changes. `DefaultAzureCredential` handles credential selection automatically based on the environment.

### Test A: Dynamic provisioning with Workload Identity ✅

AKS cluster with OIDC + WI enabled. Deployed manifests from this PR. Controller SA annotated with `azure.workload.identity/client-id`. Webhook injected all 4 env vars (`AZURE_CLIENT_ID`, `AZURE_TENANT_ID`, `AZURE_FEDERATED_TOKEN_FILE`, `AZURE_AUTHORITY_HOST`) and projected token volume. Created PVC → AMLFS created successfully (4 TiB, AMLFS-Durable-Premium-500). PVC bound.

### Test B: WI-enabled cluster without SA annotation ✅ (expected failure)

Same WI cluster, same manifests, but no `client-id` annotation on the SA. Webhook injected partial env vars (tenant + token file but not client ID). `DefaultAzureCredential` tried `WorkloadIdentityCredential` using the kubelet identity from azure.json, which has no federated credential → `AADSTS70025` error. This confirms the SA annotation is required on WI-enabled clusters.

### Test C: Managed identity only (no WI on cluster) ✅

Separate AKS cluster created **without** `--enable-workload-identity`. Deployed **upstream manifests** (no WI labels). No webhook present, no WI env vars injected. `DefaultAzureCredential` used kubelet managed identity from azure.json. Created PVC → AMLFS created successfully via MI. PVC bound. Backward compatibility confirmed.

### Test D: PSS Baseline ✅

Both clusters: controller 3/3, node (jammy) 3/3. `seccompProfile: RuntimeDefault` on pods. Sidecar containers running with `capabilities.drop: [ALL]`. Node azurelustre container retains `privileged: true` for kernel module operations.

### Key finding

On WI-enabled clusters, the pod template label `azure.workload.identity/use: "true"` causes the webhook to inject partial credentials even without the SA annotation, which breaks the MI fallback path. The updated docs now explain this: if your cluster has WI enabled, you must complete the full WI setup (annotation + federated credential) for dynamic provisioning. On clusters without WI, the labels have no effect and MI works as before.

| Test | Cluster Config | Auth Mode | Dynamic Prov | Result |
|------|---------------|-----------|-------------|--------|
| A | WI-enabled + SA annotated | WorkloadIdentityCredential | ✅ AMLFS created | PASS |
| B | WI-enabled, NO SA annotation | Partial WI injection | ❌ Auth error | EXPECTED |
| C | No WI (MI only), upstream manifests | ManagedIdentityCredential | ✅ AMLFS created | PASS |
| D | Both clusters | PSS Baseline | All pods running | PASS |

jeffbearer:
## Consideration: WI vs MI Toggleability and Helm

During testing we discovered a backward compatibility nuance worth discussing.

### The problem

The `azure.workload.identity/use: "true"` label is baked into the deploy manifests on the pod templates. On AKS clusters with workload identity enabled, this causes the mutating webhook to inject partial WI env vars into the controller pods — even when the SA `client-id` annotation is not set. This partial injection causes `DefaultAzureCredential` to attempt `WorkloadIdentityCredential` with the kubelet identity (which has no federated credential), failing before it can fall back to managed identity.

In practice, this means:

| Cluster has WI enabled? | SA annotated with client-id? | Result |
|---|---|---|
| Yes | Yes | ✅ WI auth works |
| Yes | No | ❌ Auth fails (partial injection) |
| No | N/A | ✅ MI works (webhook absent, labels inert) |

Users deploying these raw manifests on a WI-enabled cluster **must** complete the full WI setup for dynamic provisioning to work. There is no way to opt out and fall back to MI using the raw YAML manifests alone.

### Why Helm solves this

The `blob-csi-driver` handles this with Helm values:

```yaml
workloadIdentity:
  clientID: ""    # when empty, no WI labels or annotations are added
  tenantID: ""
```

When `clientID` is set, the Helm chart adds the `azure.workload.identity/use: "true"` label to pod templates and the `client-id` annotation to the SA. When empty, WI labels are omitted entirely and MI works as before. This makes the choice explicit and safe.

### Options for this PR

1. **Merge as-is for AKS extension onboarding** — The target use case (AKS extensions) requires WI anyway per the [security requirements](https://msazure.visualstudio.com/One/_wiki/wikis/One.wiki/102775/Partner-Onboarding-Process). The docs now clearly state that WI setup is required on WI-enabled clusters. Users on non-WI clusters are unaffected.

2. **Defer WI labels to Helm chart work** — Only merge the PSS hardening (seccomp, capabilities.drop) and docs in this PR. Add the WI pod template labels when the Helm chart lands (tracked under Feature 12607802), which makes it properly toggleable.

3. **Remove pod template labels, keep SA labels only** — Users would manually add the pod template label when enabling WI. Safer for backward compat but requires manual manifest editing that the docs would need to walk through.

I lean toward option 1 since AKS extension onboarding is the driving requirement, but wanted to surface this for discussion.

PR #275, opened 2026-04-08
feat: allow volume mounting of the CSI driver entrypoint script

jeffbearer:
**What type of PR is this?**

/kind feature

Adds support for overriding the built-in `entrypoint.sh` via a Kubernetes ConfigMap, enabling entrypoint changes without rebuilding the Docker image. This is useful for troubleshooting customer networking/installation issues and supporting custom initialization requirements (e.g. Nvidia).

A `start.sh` wrapper script checks for a custom entrypoint at `/app/custom-entrypoint/entrypoint.sh` (mounted from an optional ConfigMap). If found, it runs the custom version; otherwise it falls back to the built-in entrypoint. The install script gets a `--custom-entrypoint <file>` flag to create/delete the ConfigMap.

AB#34599328

**Requirements**:
- [x] includes documentation
- [ ] adds unit tests
- [x] tested upgrade from previous version

Tested on AKS (Avere_ManagedLustre_CSISmoke_MLT10_dev, canadacentral, Standard_D2s_v3, Ubuntu 22.04 jammy nodes):

| Test | Result |
|------|--------|
| Default install (no flag) | ✅ No ConfigMap, built-in entrypoint used |
| Custom entrypoint (`--custom-entrypoint`) | ✅ ConfigMap created, `start.sh` dispatched to custom, marker in logs |
| Switch custom → default (re-install without flag) | ✅ ConfigMap deleted, pods restarted, built-in restored |
| Uninstall cleanup | ✅ ConfigMap deleted with all resources |

Key finding during testing: `kubectl apply` alone does not restart pods when only the ConfigMap is created/deleted (the pod spec is unchanged). Added `kubectl rollout restart` to the install script to handle this.

**Release note**:
```
Add --custom-entrypoint flag to install-driver.sh for overriding the CSI driver entrypoint script via ConfigMap without rebuilding the Docker image.
```

### DIFF
--- a/deploy/install-driver.sh
+++ b/deploy/install-driver.sh
@@ -15,21 +15,51 @@
 # limitations under the License.
 set -euo pipefail
 
+CONFIGMAP_NAME="csi-azurelustre-entrypoint"
+
 function usage {
-    echo "Usage: $0 [branch|local|url]"
+    echo "Usage: $0 [--custom-entrypoint <file>] [branch|local|url]"
     echo
     echo "branch: The branch from which to install the Azure Lustre CSI Driver to install. Default is 'main'."
     echo "local: Deploy out of local filesystem."
+    echo
+    echo "Options:"
+    echo "  --custom-entrypoint <file>  Use a custom entrypoint script via ConfigMap instead of the"
+    echo "                              built-in entrypoint. The file will be mounted into the CSI driver"
+    echo "                              containers. Without this flag, the built-in entrypoint is used."
     echo
     echo "Example:"
     echo "$0 # install from remote main"
     echo "$0 main # install from remote branch or reference"
     echo "$0 local # install from locally checked out branch"
     echo "$0 https://raw.githubusercontent.com/csmuell/azurelustre-csi-driver/main # install from given remote repository/branch"
+    echo "$0 --custom-entrypoint ./my-entrypoint.sh local # install with custom entrypoint"
     exit 1
 }
 
-if [[ "$#" -gt 1 || ("$#" -gt 0 && "$1" == "--help") ]]; then
+custom_entrypoint=""
+
+# Parse --custom-entrypoint flag (must come before positional args)
+while [[ "$#" -gt 0 ]]; do
+  case "$1" in
+    --custom-entrypoint)
+      if [[ "$#" -lt 2 ]]; then
+        echo "Error: --custom-entrypoint requires a file path argument."
+        usage
+      fi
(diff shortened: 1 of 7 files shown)


deploy/install-driver.sh
```diff
@@ -71,6 +116,13 @@ kubectl apply -f "$repo/csi-azurelustre-controller.yaml"
 kubectl apply -f "$repo/csi-azurelustre-node-jammy.yaml"
 kubectl apply -f "$repo/csi-azurelustre-node-noble.yaml"
 
+# Restart pods to ensure they pick up ConfigMap changes.
+# kubectl apply alone won't trigger a restart when only the ConfigMap
+# was created or deleted.
+kubectl rollout restart deployment csi-azurelustre-controller -n kube-system
```
dabradley: I think in the basic case where a user isn't even considering the custom entrypoint, it would be surprising to have a restart on their deployments every time. How feasible would it be to have it only restart if the configmap is actually changing state?

jeffbearer:
Addressed all review feedback in the latest force-push (`6007ae6f`):

**@dabradley** — [restart concern](https://github.com/kubernetes-sigs/azurelustre-csi-driver/pull/275#discussion_r3059252067): Good call. The install script now tracks a `configmap_changed` flag and only restarts when the ConfigMap is actually created, updated, or deleted. Default installs (no `--custom-entrypoint` flag, no existing ConfigMap) skip the restart entirely.

**Copilot reviewer** — [controller restart](https://github.com/kubernetes-sigs/azurelustre-csi-driver/pull/275#discussion_r3059288011): Fixed. Controller restart removed since it does not mount the ConfigMap. Only node DaemonSets are restarted.

**Copilot reviewer** — [docs scope](https://github.com/kubernetes-sigs/azurelustre-csi-driver/pull/275#discussion_r3059288070): Fixed. Docs now clarify the ConfigMap is mounted into **node DaemonSet pods only** — the controller is not affected. This is intentional: the controller sets `AZURELUSTRE_CSI_INSTALL_LUSTRE_CLIENT=no` so its entrypoint is effectively a no-op passthrough. All the Lustre client installation and LNet networking that customers need to customize runs in the node entrypoint.

**Copilot reviewer** — [security note](https://github.com/kubernetes-sigs/azurelustre-csi-driver/pull/275#discussion_r3059288102): Added. Docs now include an explicit security warning about RBAC and code-injection risk for the privileged ConfigMap.

dabradley (approved):
/lgtm

PR #280, opened 2026-04-29
feat: auto-inject unique_fsid mount option for Lustre >= 2.15.8

jeffbearer:
## What type of PR is this?

/kind feature

Automatically adds the `unique_fsid` Lustre mount option when the loaded kernel module supports it (>= 2.15.8). This gives each mount its own filesystem ID so Kubernetes can properly distinguish multiple mounts of the same Lustre filesystem on a single node.

### Problem

Lustre historically forces a shared `s_dev` across all mounts of the same filesystem. Kubernetes identifies mounts by `(st_dev, fsroot)`, so it can't distinguish multiple mounts — all Lustre mounts look like the same device. This causes kubelet to block unmount of one volume while another volume's mount exists on the same node (documented in detail in the existing NodeStageVolume comment).

At scale (~110 pods/node), all pods funnel through a single Lustre client mount point, creating a performance bottleneck.

### Solution

The `unique_fsid` kernel mount option (commit c0f2040a9e in 2.15.8-client) skips the UUID-to-s_dev assignment, yielding unique FSIDs per mount. This PR:

1. **Lazy capability detection** — reads `/sys/module/lustre/version` on first mount, caches the result via `sync.Once`
2. **Auto-injection** — adds `unique_fsid` to mount options when Lustre >= 2.15.8
3. **Conservative default** — if version can't be determined, does NOT add the flag
4. **User overrides** — `no_unique_fsid` in mount options suppresses auto-injection; explicit `unique_fsid` is not duplicated
5. **Warning** — logs when user explicitly requests `unique_fsid` on an unsupported version

## Which issue(s) this PR fixes

AB#37726751

## Special notes for your reviewer

- The capability detection is on the `Driver` struct (not a package global) for testability
- `getMountOptions` is a pure function — takes `supportsUniqueFsid bool` parameter, fully testable
- `no_unique_fsid` is a CSI-driver-level sentinel, stripped before reaching the kernel
- NodeStageVolume remains unimplemented — this sets up the precondition but the staging change is a separate validation effort
- DaemonSet pod restart required after kmod upgrade for detection to refresh (documented in code)

## Requirements
- [x] uses conventional commit messages
- [x] includes documentation (updated NodeStageVolume comment)
- [x] adds unit tests

```release-note
feat: Auto-inject unique_fsid mount option for Lustre >= 2.15.8, enabling Kubernetes to distinguish multiple mounts of the same filesystem on a single node.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ Starting with v0.4.0, the driver ships separate images per Ubuntu distribution: 
 
 | Driver version | Image | Supported k8s version | Lustre client version |
 | -------------- | ----- | --------------------- | --------------------- |
-| main branch | mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.4.0-jammy<br>mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.4.0-noble | 1.21+ | 2.15.7 |
+| main branch | mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.4.0-jammy<br>mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.4.0-noble | 1.21+ | 2.15.7 (jammy) / 2.16.1 (noble) |
+| development branch | mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:latest-jammy<br>mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:latest-noble | 1.21+ | 2.15.8 (jammy) / 2.16.1 (noble) |
 | v0.4.0 | mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.4.0-jammy<br>mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.4.0-noble | 1.21+ | 2.15.7 |
 | v0.3.1 | mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.3.1 | 1.21+ | 2.15.7 |
 | v0.3.0 | mcr.microsoft.com/oss/v2/kubernetes-csi/azurelustre-csi:v0.3.0 | 1.21+ | 2.15.5 |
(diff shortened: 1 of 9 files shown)


jeffbearer:
## E2E Validation Results — unique_fsid

**Environment:**
- AKS: 2-node (Standard_D2s_v3), K8s v1.34.4, canadacentral
- AMLFS: 4 TiB Durable-Premium-500, MGS 10.1.0.7
- Lustre kmod: 2.15.8_34_gc0f2040
- CSI image: built from this branch

### Test 1: Auto-injection
Created PV with standard mount options (noatime, flock). No explicit unique_fsid.
`/proc/mounts` on the node shows unique_fsid was auto-injected:
```
10.1.0.7@tcp:/lustrefs .../mount lustre rw,noatime,...,unique_fsid,encrypt 0 0
```
**PASS** — capability detection identified 2.15.8, injected automatically.

### Test 2: Unique f_fsid per volume
Two PVs → same AMLFS, different sub-dirs, same node.
```
Pod 1: f_fsid = 10000400000000
Pod 2: f_fsid = 1000e700000000
```
**PASS** — different IDs. K8s can distinguish them.

### Test 3: no_unique_fsid opt-out
Two PVs with `no_unique_fsid` in mountOptions, same node.
```
Pod 3: f_fsid = 481510a200000000
Pod 4: f_fsid = 481510a200000000
```
Kernel mount shows `nounique_fsid` (not our sentinel `no_unique_fsid`) — sentinel correctly stripped.
**PASS** — same f_fsid, old shared behavior restored.

### Test 4: Unmount isolation
4 volumes on one node (2 with unique_fsid, 2 without). Deleted Pod 2 + Pod 4.
- Mounts for deleted volumes removed
- Pod 1 and Pod 3 unaffected (data accessible)

**PASS** — unmounting one volume does not affect others.

### Test 5: Multi-pod shared volume (RWX)
Two pods sharing one PVC on the same node with unique_fsid active.
```
Pod B reads Pod A file: "hello from pod-a" ✅
Pod A reads Pod B file: "reply from pod-b" ✅
```
**PASS** — data sharing works correctly.

### Summary
| Test | Result |
|------|--------|
| Auto-injection (>= 2.15.8) | ✅ PASS |
| Unique f_fsid per volume | ✅ PASS |
| no_unique_fsid opt-out | ✅ PASS |
| Sentinel stripping | ✅ PASS |
| Unmount isolation | ✅ PASS |
| Multi-pod shared volume | ✅ PASS |

PR #297, opened 2026-05-29
Enable shellcheck

dabradley:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


test/sanity/run-tests-all-clouds.sh
```diff
@@ -14,14 +14,13 @@
...
+set -euo pipefail
 
+# shellcheck disable=SC2154 # GOPATH: standard Go environment variable, expected to be set externally
 function install_csi_sanity_bin {
   echo 'Installing CSI sanity test binary...'
-  mkdir -p $GOPATH/src/github.com/kubernetes-csi
-  pushd $GOPATH/src/github.com/kubernetes-csi
+  mkdir -p "${GOPATH}"/src/github.com/kubernetes-csi
+  pushd "${GOPATH}"/src/github.com/kubernetes-csi
   export GO111MODULE=off
```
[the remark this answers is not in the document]
dabradley: This test isn't even used right now. It will probably be replaced or changed with upcoming testing changes. Not fixing now.

hack/verify-shellcheck.sh
```diff
@@ -0,0 +1,72 @@
...
+    echo "Every inline disable must have a trailing comment, e.g.:" >&2
+    echo "  # shellcheck disable=SC2154 # VAR is set by the calling script" >&2
+    echo "" >&2
+    for line in "${bare_disables[@]}"; do
+        echo "  ${line}" >&2
+    done
+    exit 1
+fi
+
+shellcheck -x -o all -S style "${scripts[@]}"
```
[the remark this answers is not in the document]
dabradley: This is only supported in shellcheck 0.10.0, which isn't even available in my noble install. Not worth breaking the script for many devs over. Alternatives are cumbersome and unnecessary since this is a theoretical edge-case concern

hack/verify-yamllint.sh
```diff
@@ -14,41 +14,25 @@
...
-    if [ $linecount -gt 0 ]; then
-        echo "yaml files under $path are not linted, failed with: "
-        cat $LOG
+    echo "checking yamllint under path: ${path} ..."
+    yamllint -f parsable "${path}" > "${LOG}"
+    linecount=$(wc -l < "${LOG}")
+    if [[ "${linecount}" -gt 0 ]]; then
+        echo "yaml files under ${path} are not linted, failed with: "
+        cat "${LOG}"
         exit 1
```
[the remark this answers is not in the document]
dabradley: Fixed by improving script

hack/verify-shellcheck.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# Require every inline `# shellcheck disable=` to have a trailing comment
+# explaining why the rule is suppressed.  shellcheck itself has no flag for
+# this, so we enforce it with a simple grep.
+bare_disables=()
+for script in "${scripts[@]}"; do
+    while IFS= read -r match; do
+        bare_disables+=("${match}")
+    done < <(grep -nP '^\s*#\s*shellcheck\s+disable=SC[0-9]+(,SC[0-9]+)*\s*$' "${script}" | sed "s|^|${script}:|")
+done
```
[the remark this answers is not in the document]
dabradley: Fixed by improving script

hack/verify-helm-chart.sh
```diff
@@ -43,18 +44,16 @@ if [[ -z "$(command -v pip)" ]]; then
...
-  echo "Cannot find jq. Installing yq..."
-  apt install jq -y
-fi
-
 # jq-equivalent for yaml
-pip install yq
+if [[ -z "$(command -v yq)" ]]; then
+  echo "Cannot find yq. Installing yq..."
+  apt install yq -y
+fi
```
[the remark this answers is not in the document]
dabradley: This entire file is removed in a later change. No need to fix this

test/long-haul/external-e2e.sh
```diff
@@ -19,15 +19,20 @@ set -o pipefail
...
 
 echo "Downloading kubectl ${kubernetesVersion}"
-curl -Lo ${REPO_ROOT_PATH}/kubectl "https://dl.k8s.io/release/${kubernetesVersion}/bin/linux/amd64/kubectl"
-chmod a+x ${REPO_ROOT_PATH}/kubectl
-export PATH=$(pwd):${PATH}
+# shellcheck disable=SC2154 # REPO_ROOT_PATH: expected env var exported by calling script (start-long-haul.sh)
+curl -fLo "${REPO_ROOT_PATH}/kubectl" "https://dl.k8s.io/release/${kubernetesVersion}/bin/linux/amd64/kubectl"
+chmod a+x "${REPO_ROOT_PATH}/kubectl"
+PATH="$(pwd):${PATH}"
+export PATH
```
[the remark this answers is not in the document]
dabradley: Fixed

dabradley:
@jeffbearer Did the copilot back and forth and this looks pretty good now

hack/verify-shellcheck.sh
```diff
@@ -0,0 +1,71 @@
...
+echo "Verifying ${#scripts[@]} shell scripts with shellcheck -x -o all -S style ..."
+
+# Require every inline `# shellcheck disable=` to have a trailing comment
+# explaining why the rule is suppressed.  shellcheck itself has no flag for
+# this, so we enforce it with grep: the regex matches a disable directive
+# whose remainder (after `=`) contains no `#`, i.e. no justification comment.
+# `grep -H` prefixes each match with the file path, giving us a ready-to-print
+# `file:line:directive` line.  `|| true` keeps us from tripping `set -e` when
+# nothing matches (grep exits 1).
+mapfile -t bare_disables < <(grep -HnE '^[[:space:]]*# shellcheck disable=[^#]*$' "${scripts[@]}" || true)
```
jeffbearer: The ^[[:space:]]* anchor means this only catches # shellcheck disable=... directives that start at the beginning of a line. Trailing-comment directives like

  some_command --foo # shellcheck disable=SC2086

 bypass the rationale gate entirely. There are no offenders today, but it's a load-bearing piece of policy machinery worth tightening.

 Suggestion — drop the start-of-line anchor and require a # justification after the directive payload:

  mapfile -t bare_disables < <(grep -HnE '# shellcheck disable=[A-Z0-9, ]+[[:space:]]*$' "${scripts[@]}" || true)

 (The trailing $ and absence of # after the SC codes is what flags "no justification". Both end-of-line and line-leading variants are now caught.)

hack/verify-shellcheck.sh
```diff
@@ -0,0 +1,71 @@
...
+           -path "${PKG_ROOT}/vendor" -o \
+           -path "${PKG_ROOT}/.jj" -o \
+           -path "${PKG_ROOT}/.git" \) -prune -o \
+        -type f \( -name '*.sh' -o -name '*.bash' \) -print \
+    | sort
+)
+
+if [[ "${#scripts[@]}" -eq 0 ]]; then
+    echo "No shell scripts found to lint."
+    exit 0
```
jeffbearer: If the find above ever returns zero matches — a rename, a refactor that moves scripts out of the tree, or a CI job that happens to run from the wrong working directory — this prints a friendly message and exits 0. Discovery regression then looks identical to a healthy "no scripts" state, and the gate silently stops protecting the repo.

 Given the repo unambiguously does have shell scripts (and is going to grow more, not fewer), I'd flip this to fail-closed:

  if [[ "${#scripts[@]}" -eq 0 ]]; then
      echo "ERROR: no shell scripts discovered under ${PKG_ROOT}; check the find filter and prune list." >&2
      exit 1
  fi

 If there's ever a legitimate reason to allow zero matches (e.g. running against an external checkout that genuinely has none), gating on an env var like ALLOW_EMPTY=1 would preserve the safety net.

hack/verify-shellcheck.sh
```diff
@@ -0,0 +1,71 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+PKG_ROOT=$(git rev-parse --show-toplevel)
+
+if [[ -z "$(command -v shellcheck)" ]]; then
```
jeffbearer: This check confirms shellcheck is on PATH but not what version it is. Two concrete examples of why that matters: on shellcheck 0.11.0 (what apt install shellcheck gives me locally), two additional findings fire that CI's older shellcheck doesn't:

  - SC2335 in hack/verify-helm-chart.sh:33 (! a == b → a != b)
  - SC2329 in hack/verify-spelling.sh:32, which may also make the existing # shellcheck disable=SC2317 directive on line 31 stale (newer shellcheck moved that diagnostic to SC2329)

 These will surface the next time the ubuntu-latest runner image bumps its packaged shellcheck — at which point a previously-green PR breaks CI for unrelated reasons.

 Two low-effort options:

  1. Pin in CI — if shellcheck is being installed via apt in the workflow, switch to ludeeus/action-shellcheck with a pinned version:, and document it in this script's header.
  2. Floor-check here — assert a minimum version inline so local devs and CI fail loudly when they're below it:
  min_version="0.10.0"
  actual_version=$(shellcheck --version | awk '/^version:/ {print $2}')
  if [[ "$(printf '%s\n%s\n' "${min_version}" "${actual_version}" | sort -V | head -n1)" != "${min_version}" ]]; then
      echo "shellcheck >= ${min_version} required; found ${actual_version}." >&2
      exit 1
  fi

 Either is fine; the goal is just to make the contract on shellcheck's version explicit instead of implicit.

hack/verify-yamllint.sh
```diff
@@ -14,41 +14,21 @@
...
 do
-    echo "checking yamllint under path: $path ..."
-    yamllint -f parsable $path > $LOG
-    cat $LOG
-    linecount=`cat $LOG | wc -l`
-    if [ $linecount -gt 0 ]; then
-        echo "yaml files under $path are not linted, failed with: "
-        cat $LOG
+    echo "checking yamllint under path: ${path} ..."
+    if ! output=$(yamllint -f parsable "${path}" 2>&1); then
```
jeffbearer: Small regression vector compared to the pre-PR script (which failed on any parseable output, not just non-zero exit). yamllint returns exit 0 when the only diagnostics emitted are severity: warning — so any future rule added to .yamllint.yaml with warning severity will print findings to the log and let CI pass green.

 There are no warning-severity rules in .yamllint.yaml today, so this is latent — but the whole point of this PR is to make the verify scripts trustworthy, and this one currently isn't, for warnings.

fix:

      if ! output=$(yamllint --strict -f parsable "${path}" 2>&1); then

 --strict promotes warnings to errors at the exit-code level. Behavior is identical for the current ruleset (no warnings exist), and the file gains a real guarantee.

hack/verify-helm-chart.sh
```diff
@@ -43,18 +44,16 @@ if [[ -z "$(command -v pip)" ]]; then
...
-if [[ -z "$(command -v jq)" ]]; then
-  echo "Cannot find jq. Installing yq..."
-  apt install jq -y
-fi
-
 # jq-equivalent for yaml
-pip install yq
+if [[ -z "$(command -v yq)" ]]; then
+  echo "Cannot find yq. Installing yq..."
+  apt install yq -y
```
jeffbearer: This is a real toolchain swap, not a refactor: the old pip install yq pulled kislyuk/yq (a Python wrapper around jq), and apt install yq pulls mikefarah/yq (the Go-based standalone with a different parser and CLI surface). The simple path queries in this script (.image.X.repository, .spec.template.spec.containers[N].image) should work on both, but a couple of things worth verifying before this script is un-commented in hack/verify-all.sh:

  1. -r flag semantics. In kislyuk/yq, -r is jq's "raw output" (strips surrounding quotes from string scalars). In mikefarah/yq v4, scalar leaves print unquoted by default and -r isn't a standard flag — depending on the version apt ships, it may be silently ignored, aliased, or rejected. Worth a one-line manual test on a CI-like runner to confirm yq -r .image.csiProvisioner.repository charts/.../values.yaml returns a bare string and not "bare-string". Lines 24, 25,54-56, 68-70 all rely on this.
  2. The pip install block above (lines 41-45) is now dead code — the new yq doesn't need pip. Worth deleting in the same commit as the toolchain swap so it's clear they go together.
  3. Blast radius is currently zero because verify-all.sh has this script commented out (lines 29-30), so any breakage here won't be caught by CI — making point 1 above more important, not less.

jeffbearer (changes requested):
Take a look at the comments and if they are in scope.

hack/verify-shellcheck.sh
dabradley: That's legit. I'll fix that

hack/verify-shellcheck.sh
dabradley: Valid point. I like that. I have some plans on improving a lot of these scripts, but that's reasonable to update now

hack/verify-shellcheck.sh
dabradley: Yeah, I was going to go over these scripts to make sure we're keeping versions in line. I'll take a look at this and see if I can prevent this now before we hit something like that.

hack/verify-yamllint.sh
dabradley: I'm never a fan of warnings (easy to ignore then they build), so having an explicit --strict is probably the best here.

hack/verify-helm-chart.sh
dabradley: This whole script goes away with the charts changes. I have written three more targeted helm chart checks for that. I probably should have done the least effort to shut up shellcheck here (probably just deleted it honestly). Maybe I'll just delete it if my excuse is 'it's going away later'.

jeffbearer (approved):
/lgtm
/approve

PR #300, opened 2026-06-24
Add charts

dabradley:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


deploy/csi-azurelustre-controller.yaml
```diff
@@ -4,39 +4,50 @@ apiVersion: apps/v1
...
+    app.kubernetes.io/name: azurelustre-csi-driver
+    app.kubernetes.io/part-of: azurelustre-csi-driver
+    app.kubernetes.io/version: latest
+    app: csi-azurelustre-controller
 spec:
   replicas: 2
   selector:
     matchLabels:
       app: csi-azurelustre-controller
+      app.kubernetes.io/component: controller
```
jeffbearer, line 18: **Heads-up: this `app.kubernetes.io/component` lands in `spec.selector.matchLabels`, which is immutable on a Deployment.**

Because the controller keeps its name across this change, an in-place `kubectl apply` upgrade — the path `deploy/install-driver.sh` uses — fails here:

```
The Deployment "csi-azurelustre-controller" is invalid: spec.selector: ... field is immutable
```

I reproduced this against a live API server: the base manifest applies, then applying this version over it errors on the immutable selector, while a fresh install succeeds — so it's strictly an upgrade-path issue. `install-driver.sh` already deletes the renamed node DaemonSet and RBAC objects, but the controller isn't renamed or deleted, so it hits this.

Since `app: csi-azurelustre-controller` already uniquely selects the controller pods, we can keep `component` in `metadata.labels` and `spec.template.metadata.labels` (both mutable) and leave it out of the selector:

```suggestion
      app: csi-azurelustre-controller
```

The same line should come out of `templates/controller-deployment.yaml` to keep the deploy↔chart parity check green. The node DaemonSet selectors picked up the same label too — those don't break (the DaemonSet is renamed, so the old one is removed first), but trimming them for consistency would be a nice follow-up.

charts/latest/azurelustre-csi-driver/templates/NOTES.txt
```diff
@@ -0,0 +1,10 @@
+Azure Lustre CSI Driver (latest) installed.
+
+To verify:
+  kubectl -n kube-system get pods -l app.kubernetes.io/instance={{ .Release.Name }} -o wide
+
+CSIDriver object:
+  kubectl get csidriver {{ .Values.csidriver.name }} -o yaml
+
+Uninstall:
+  helm uninstall {{ .Release.Name }} -n kube-system
```
jeffbearer: **Small polish: these helper commands hardcode `kube-system`, but the chart installs into `.Release.Namespace`.**

Everything the chart renders (Deployments, DaemonSets, ServiceAccounts) already uses `{{ .Release.Namespace }}`, so if someone installs into a different namespace this post-install output points them at the wrong place. Using the release namespace here keeps the hints accurate for every install:

```suggestion
  kubectl -n {{ .Release.Namespace }} get pods -l app.kubernetes.io/instance={{ .Release.Name }} -o wide

CSIDriver object:
  kubectl get csidriver {{ .Values.csidriver.name }} -o yaml

Uninstall:
  helm uninstall {{ .Release.Name }} -n {{ .Release.Namespace }}
```

charts/README.md
```diff
@@ -1,5 +1,113 @@
...
+```console
+helm uninstall azurelustre -n kube-system
+```
+
+## Tips
+
+- Dry run rendering: `helm template test ./charts/latest/azurelustre-csi-driver -n kube-system | less`
+- Skip Lustre client install on nodes: `--set node.lustreClient.install=false`
+- Change Lustre client version (per flavor): `--set node.jammy.lustreClient.version=2.15.7 --set node.jammy.lustreClient.shaSuffix=<sha>` (similarly for `node.noble`)
+- Increase verbosity: `--set controller.extraArgs={"-v=5"} --set node.extraArgs={"-v=5"}`
```
jeffbearer: **Heads-up: this tip silently drops a production-safety default.**

`extraArgs` is a full list replacement, so `--set controller.extraArgs={"-v=5"}` removes the chart's default `--enable-azurelustre-mock-dyn-prov=false`. I verified what that means against the compiled driver:

- `azurelustre --help` → `-enable-azurelustre-mock-dyn-prov ... (default true)` (Go prints `(default …)` only for non-false bools — the adjacent `-enable-azurelustre-mock-mount` shows no default).
- Run with the flag omitted and no cloud config present → `azurelustre.go:207 no cloud config provided, driver running with mock dynamic provisioning`.
- Run with `--enable-azurelustre-mock-dyn-prov=false` → `azurelustre.go:211 no cloud config provided, error` (fails loudly, as intended).

So dropping the flag re-enables a test-only mock provisioner in production — the code comment at `azurelustre.go:146` says "DO NOT set as true in non-testing scenario".

On verbosity: `-v=5` is already the default in `values.yaml`, and the driver's own `klog.V(...)` calls top out at level 5 (levels in use: 2, 4, 5 — no `V(6)+`), so this example doesn't actually raise verbosity. If you override `controller.extraArgs`, the key point is to keep the safety flag:

```suggestion
- Override controller args (keep `--enable-azurelustre-mock-dyn-prov=false`, or the controller falls back to a test-only mock provisioner): `--set "controller.extraArgs={-v=5,--enable-azurelustre-mock-dyn-prov=false}"`
- Override node args: `--set "node.extraArgs={-v=5}"`
```

docs/install-csi-driver.md
```diff
@@ -4,6 +4,31 @@ This document explains how to install Azure Lustre CSI driver on a kubernetes cl
 
 ## Instructions for current production release
 
+### Install with Helm (recommended)
+
+Helm is the recommended installation method for production clusters.
```
jeffbearer, line 9: **Since this section now recommends Helm, it's worth a short migration caveat for existing clusters.**

Helm only manages resources it created. On a cluster already running the `kubectl` / `install-driver.sh` path, the cluster-scoped objects — the `azurelustre.csi.azure.com` CSIDriver and the `csi-azurelustre-*` ClusterRoles — have no Helm ownership metadata, so `helm install` aborts. I verified this against a live API server:

```
Error: INSTALLATION FAILED: Unable to continue with install: CSIDriver "azurelustre.csi.azure.com" ... exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name" ...
```

Consider adding a note that users migrating from the kubectl install should remove the existing objects first (e.g. `deploy/uninstall-driver.sh`) before `helm install`, so the recommendation doesn't surprise anyone upgrading in place.

charts/latest/azurelustre-csi-driver/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+appVersion: latest
+description: Azure Lustre CSI Driver Helm chart
```
jeffbearer, line 3: **Optional polish: a couple of recommended `Chart.yaml` metadata fields are missing.**

`helm lint` already notes `[INFO] Chart.yaml: icon is recommended`. For a chart that gets published to the repo's Helm index, adding `home` and `sources` improves the rendered listing (Artifact Hub, `helm show chart`) and helps users find the project:

```suggestion
description: Azure Lustre CSI Driver Helm chart
home: https://github.com/kubernetes-sigs/azurelustre-csi-driver
sources:
  - https://github.com/kubernetes-sigs/azurelustre-csi-driver
```

`maintainers` and an `icon` are also recommended — left out of the suggestion since they need real values (owner handles and a logo URL).

deploy/csi-azurelustre-controller.yaml
dabradley: The app.kubernetes.io/component label is part of Kubernetes’ recommended labeling convention. I saw this in an earlier copilot review, and considering our recommended upgrade path is to uninstall then reinstall, and apparently everyone installs from some helm charts in some other repo anyway I didn't think it was a big deal but technically I guess we don't demand anyone do that upgrade process, so I'll just add a line to delete any existing controller before installing the new one in install-driver.sh

charts/README.md
dabradley: I just flipped the default to enable dynprov. Clearly it's working in production.

charts/latest/azurelustre-csi-driver/Chart.yaml
dabradley: Added

docs/install-csi-driver.md
dabradley: Added

charts/latest/azurelustre-csi-driver/templates/NOTES.txt
dabradley: Fixed

jeffbearer (approved):
/lgtm
/approve

PR #303, opened 2026-06-29
Chart updates for extension requirements

dabradley:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


deploy/pdb-csi-azurelustre-controller.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: csi-azurelustre-controller
+  namespace: kube-system
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/name: azurelustre-csi-driver
+    app.kubernetes.io/part-of: azurelustre-csi-driver
+    app.kubernetes.io/version: latest
+spec:
+  minAvailable: 1
```
jeffbearer: `minAvailable: 1` protects a 2-replica controller correctly, but if the controller is ever scaled to a single replica it requires zero voluntary evictions, so node drains and upgrades never complete ([k8s docs](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)). `maxUnavailable: 1` gives the same "only one controller down at a time" guarantee at 2 replicas and stays safe at 1 replica, and per the Kubernetes docs it "is recommended as it automatically responds to changes in the number of replicas of the corresponding controller":

```suggestion
  maxUnavailable: 1
```

Note: the Helm template `templates/controller-pdb.yaml` also uses `minAvailable: 1`, so it needs the same change to keep `hack/verify-helm-chart-files.sh` green (it renders the template at the default 2 replicas and diffs it against this file). With `maxUnavailable`, the template's `if replicas > 1` guard is no longer needed since the budget is safe at any replica count.

dabradley: Changed this to maxUnavailable: 1 (which is fine since we never really need more than 2), and also reconsidered how we're doing the node configuration, and changed our node value to 10% to avoid overly-long upgrades if the user has many nodes (per the onboarding docs)

jeffbearer (approved):
/lgtm
/approve

PR #307, opened 2026-07-15
Enable actionlint

dabradley:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


hack/verify-actionlint.sh
```diff
@@ -0,0 +1,51 @@
...
+}
+trap exitHandler EXIT
+
+if [[ -z "$(command -v actionlint)" ]]; then
+  echo "Cannot find actionlint. Installing actionlint..."
+  # perform go install in a temp dir as we are not tracking this version in a go module
+  # if we do the go install in the repo, it will create / update a go.mod and go.sum
+  cd "${TMP_DIR}"
+  GO111MODULE=on GOBIN="${TMP_DIR}" go install "github.com/rhysd/actionlint/cmd/actionlint@${TOOL_VERSION}"
+  export PATH="${TMP_DIR}:${PATH}"
```
csmuell, line 43: i don't think we're preferencing the pinned shellcheck; so could get inconsistencies w/ system packages.

maybe add something like

```
shopt -s nullglob; sc=(/tmp/shellcheck-v*/)
export PATH="${sc:+${sc[0]%/}:}${PATH}"
```

here before we run actionlint.

PR #311, opened 2026-07-27
Workload identity support

dabradley:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azurelustre/azurelustre.go
```diff
@@ -240,6 +241,21 @@ func NewDriver(options *DriverOptions) (*Driver, error) {
...
+			// operators can see which credential DefaultAzureCredential selected,
+			// unless the operator opted into full SDK logging via
+			// AZURE_SDK_GO_LOGGING, whose verbose request/response output can
+			// contain sensitive data and should not be forwarded automatically.
+			if os.Getenv("AZURE_SDK_GO_LOGGING") == "" {
+				azlog.SetEvents(azidentity.EventAuthentication)
+				azlog.SetListener(func(event azlog.Event, msg string) {
+					klog.V(2).Infof("azidentity %s: %s", event, msg)
+				})
+			}
```
jeffbearer: Minor, non-blocking: this gate keys off "non-empty" (`== ""`), but azcore only enables its own logging when the value is **exactly** `"all"` (vendored `sdk/internal/log/log.go` `init()`). That mismatch creates a dead zone: a value like `AZURE_SDK_GO_LOGGING=true` makes this block back off (it assumes the operator opted into full logging) while azcore never turns on (the value isn't `"all"`) — so **nothing logs at all**, not even the auth events. That's strictly less observability than the default (unset) case.

| `AZURE_SDK_GO_LOGGING` | azcore `init()` | this block | net result |
|---|---|---|---|
| unset | no listener | installs auth-only klog listener | ✅ auth events → klog `V(2)` |
| `"all"` | stderr listener (all events) | backs off | ✅ full verbose SDK logging |
| `"true"` / `"1"` / … | **no listener** | backs off | ❌ nothing logs |

Suggest matching azcore's own predicate (`!= "all"`) so only an exact `"all"` — which azcore honors with full stderr logging — makes the driver step aside; every other value still gets auth-event forwarding. This also tweaks the comment to name the exact value so the non-empty assumption isn't reintroduced later.

```suggestion
			// Controller pod: make the Azure identity used for ARM calls
			// observable. Forward only azidentity Authentication events to klog so
			// operators can see which credential DefaultAzureCredential selected,
			// unless the operator opted into full SDK logging by setting
			// AZURE_SDK_GO_LOGGING=all, whose verbose request/response output can
			// contain sensitive data and should not be forwarded automatically.
			if os.Getenv("AZURE_SDK_GO_LOGGING") != "all" {
				azlog.SetEvents(azidentity.EventAuthentication)
				azlog.SetListener(func(event azlog.Event, msg string) {
					klog.V(2).Infof("azidentity %s: %s", event, msg)
				})
			}
```

(Unrelated aside: forwarded events land at `klog.V(2)`, so they only surface at `-v=2`+ — worth confirming the controller's default verbosity shows them.)

docs/csi-debug.md
```diff
@@ -836,6 +836,41 @@ kubectl logs -n kube-system -l app=csi-azurelustre-controller -c azurelustre --t
...
+# The controller pods should have the injected workload-identity env vars
+kubectl exec -n kube-system deploy/csi-azurelustre-controller -c azurelustre -- env | grep AZURE_
+```
+
+Expect `AZURE_CLIENT_ID`, `AZURE_TENANT_ID`, and `AZURE_FEDERATED_TOKEN_FILE` to all be present. If `AZURE_CLIENT_ID` is missing while the other two are set, the ServiceAccount is missing its `azure.workload.identity/client-id` annotation.
+
+Confirm which credential served the Azure token:
+
+```sh
+kubectl logs -n kube-system deploy/csi-azurelustre-controller -c azurelustre | grep "authenticated with"
```
jeffbearer: Nit / consistency: these two new commands reference the deployment by name (`deploy/csi-azurelustre-controller`), which only resolves when the chart is installed with `fullnameOverride=csi-azurelustre` — the deployment is otherwise named `{{ include "azurelustre.fullname" . }}-controller`. Every other command in this doc uses the release-agnostic selector `-l app=csi-azurelustre-controller`, so these will surprise anyone on a different release name.

For the `logs` command, switching to the selector form matches the rest of the file (and its `--tail=300`/`grep -i` style):

```suggestion
kubectl logs -n kube-system -l app=csi-azurelustre-controller -c azurelustre --tail=300 | grep -i "authenticated with"
```

The `kubectl exec ... deploy/csi-azurelustre-controller` line above (env-var check) has the same name assumption, but `exec` doesn't accept `-l`, so it can't use a selector directly. Suggest resolving a pod first, e.g.:

```sh
POD=$(kubectl get pod -n kube-system -l app=csi-azurelustre-controller -o jsonpath='{.items[0].metadata.name}')
kubectl exec -n kube-system "$POD" -c azurelustre -- env | grep AZURE_
```

charts/latest/azurelustre-csi-driver/templates/controller-deployment.yaml
```diff
@@ -1,3 +1,6 @@
+{{- if eq .Values.IsWorkloadIdentityEnabled "Enabled" -}}
+{{- $_ := required "IdentityClientId must be set when IsWorkloadIdentityEnabled is Enabled" .Values.IdentityClientId -}}
+{{- end -}}
```
jeffbearer, line 6: **Silent auth misdirection when `serviceAccount.create=false`.** The pod label `azure.workload.identity/use: "true"` is rendered here whenever `IsWorkloadIdentityEnabled == "Enabled"`, but the `azure.workload.identity/client-id` annotation is only emitted by `controller-serviceaccount.yaml`, which is gated on `{{- if .Values.serviceAccount.create }}`. So with a bring-your-own SA (`serviceAccount.create=false`) + WI enabled:

- the webhook still projects the federated-token volume (label is present), but
- the SA carries no `client-id` annotation → `AZURE_CLIENT_ID` is never injected → `NewDefaultAzureCredential` **silently falls back to the node/kubelet managed identity** instead of failing.

The existing `required IdentityClientId` guard doesn't catch this — it validates the value, but the SA template that would consume it is skipped. Result: dynamic provisioning authenticates as the *wrong* identity with no signal.

Suggest turning this into a render-time failure so the misconfiguration can't ship silently:

```suggestion
{{- if eq .Values.IsWorkloadIdentityEnabled "Enabled" -}}
{{- $_ := required "IdentityClientId must be set when IsWorkloadIdentityEnabled is Enabled" .Values.IdentityClientId -}}
{{- if not .Values.serviceAccount.create -}}
{{- fail "IsWorkloadIdentityEnabled is Enabled but serviceAccount.create is false. The chart adds the azure.workload.identity/client-id annotation only to a ServiceAccount it manages, so the controller would silently fall back to the node identity. Set serviceAccount.create=true, or add the annotation to your ServiceAccount out-of-band." -}}
{{- end -}}
{{- end -}}
```

Tradeoff worth your call: this hard-fails the BYO-SA + WI combination even for an operator who *did* annotate their own SA correctly, because the chart can't verify an unmanaged SA's annotations. If you want to keep that path open, an alternative is an escape-hatch value (e.g. `serviceAccount.externallyAnnotated: true`) that suppresses the `fail`. Either way, failing loudly is strictly safer than the current silent fallback to the wrong identity.

pkg/azurelustre/dynamic_provisioning.go
```diff
@@ -45,6 +45,7 @@ const (
 	AmlfsSkuResourceType                       = "amlFilesystems"
 	AmlfsSkuCapacityIncrementName              = "OSS capacity increment (TiB)"
 	AmlfsSkuCapacityMaximumName                = "default maximum capacity (TiB)"
+	authFailureHintFormat                      = "error occurred calling API: %v; verify controller identity RBAC and workload identity configuration"
```
jeffbearer: **The auth hint assumes workload identity even when it's disabled (the default).** This constant is consumed by `convertHTTPResponseErrorToGrpcCodeError` (lines 103/105), which is the shared 401/403 mapper used by ~15 ARM call sites (create, delete, subnet-capacity, SKU listing, retry checks). Every ARM `403`/`401` — including on clusters running with the default `IsWorkloadIdentityEnabled: "Disabled"` (managed/kubelet identity) — now tells the operator to "verify ... workload identity configuration," which is misleading when no workload identity is in play.

Minimal fix: keep the hint in the shared mapper but make the workload-identity part conditional in the wording, so it's accurate for both identity modes:

```suggestion
	authFailureHintFormat                      = "error occurred calling API: %v; verify the controller identity has the required RBAC role assignments (for workload identity, confirm its federated credential is configured)"
```

Note: this reword will break three assertions in `dynamic_provisioning_test.go` that match the old substring — lines **1583** (`expectedError`), **1635** and **1641** (`expectedMsgContains`). They'd need updating to a stable substring of the new message, e.g. `"required RBAC role assignments"`.

(Larger, optional alternative if you'd rather not carry identity vocabulary in a generic HTTP→gRPC mapper at all: have the mapper return the bare `PermissionDenied`/`Unauthenticated`, and let the provisioning-operation layer — which knows the identity mode — attach the remediation text. More invasive; the reword above is the low-risk version.)

charts/latest/azurelustre-csi-driver/templates/controller-deployment.yaml
```diff
@@ -1,3 +1,6 @@
+{{- if eq .Values.IsWorkloadIdentityEnabled "Enabled" -}}
```
jeffbearer, line 4: **`IsWorkloadIdentityEnabled` is matched with an exact-string `eq`, but nothing validates the value — a typo silently disables the whole feature *and* bypasses the `IdentityClientId` guard.** Because the `required IdentityClientId` check (line 2) is itself nested inside `if eq .Values.IsWorkloadIdentityEnabled "Enabled"`, any value that isn't exactly `Enabled` makes both the guard and the label/annotation rendering no-ops.

Trigger: `--set IsWorkloadIdentityEnabled=enabled` (lowercase), `=true` (YAML bool), or `="Enabled "` (trailing space). The operator believes workload identity is on; the controller comes up healthy; then every ARM call fails with `403` → mapped to `PermissionDenied`, which external-provisioner retries indefinitely. There's no allowed-value enforcement anywhere.

Suggest a `fail`-based allowed-value guard at the top of the file so a bad value is rejected at render time instead of failing silently at provisioning time:

```suggestion
{{- if not (has .Values.IsWorkloadIdentityEnabled (list "Enabled" "Disabled")) -}}
{{- fail (printf "IsWorkloadIdentityEnabled must be \"Enabled\" or \"Disabled\", got: %v" .Values.IsWorkloadIdentityEnabled) -}}
{{- end -}}
{{- if eq .Values.IsWorkloadIdentityEnabled "Enabled" -}}
```

This composes with the `serviceAccount.create` guard I suggested just below (they're independent checks in the same top block). The same value is also compared in `controller-serviceaccount.yaml`; centralizing all three `eq ... "Enabled"` comparisons behind a `define "azurelustre.workloadIdentityEnabled"` helper would be a nice follow-up, but the guard above is the safety-critical part.

hack/verify-helm-chart-files.sh
```diff
@@ -461,6 +461,54 @@ check_conditional_blocks() {
...
+    --set "IsWorkloadIdentityEnabled=Enabled" --set "IdentityClientId=test-client-id" \
+    --set "IdentityTenantId=test-tenant-id" --namespace kube-system \
+    chart-test "${chart_dir}" 2>&1); then
+    echo "ERROR: helm template failed with workload identity enabled:"
+    echo "${rendered}"
+    return 1
+  fi
+  pod_label=$(printf '%s\n' "${rendered}" | yq eval 'select(.kind == "Deployment") | .spec.template.metadata.labels."azure.workload.identity/use"' -)
+  client_id=$(printf '%s\n' "${rendered}" | yq eval 'select(.kind == "ServiceAccount" and .metadata.name == "csi-azurelustre-controller-sa") | .metadata.annotations."azure.workload.identity/client-id"' -)
+  tenant_id=$(printf '%s\n' "${rendered}" | yq eval 'select(.kind == "ServiceAccount" and .metadata.name == "csi-azurelustre-controller-sa") | .metadata.annotations."azure.workload.identity/tenant-id"' -)
```
jeffbearer: **Single-source-of-truth: the CI check re-hardcodes the SA name instead of deriving it from the chart.** The literal `csi-azurelustre-controller-sa` is authored in at least six places — the `azurelustre.serviceAccountNameController` define in `_helpers.tpl`, the federated-credential `--subject` in the docs (×3), and these two yq selectors. Because this check hardcodes the same literal it's supposed to be verifying, it can't catch a drift between the controller Deployment's `serviceAccountName` and the annotated ServiceAccount — the exact invariant workload identity depends on (webhook injects `AZURE_CLIENT_ID` only if the SA the pod runs as carries the `client-id` annotation).

Suggest deriving the SA name from the rendered Deployment so the check validates the chart's actual wiring and stays correct if the default name ever changes:

```suggestion
  local sa_name
  sa_name=$(printf '%s\n' "${rendered}" | yq eval 'select(.kind == "Deployment") | .spec.template.spec.serviceAccountName' -)
  pod_label=$(printf '%s\n' "${rendered}" | yq eval 'select(.kind == "Deployment") | .spec.template.metadata.labels."azure.workload.identity/use"' -)
  client_id=$(printf '%s\n' "${rendered}" | yq eval "select(.kind == \"ServiceAccount\" and .metadata.name == \"${sa_name}\") | .metadata.annotations.\"azure.workload.identity/client-id\"" -)
  tenant_id=$(printf '%s\n' "${rendered}" | yq eval "select(.kind == \"ServiceAccount\" and .metadata.name == \"${sa_name}\") | .metadata.annotations.\"azure.workload.identity/tenant-id\"" -)
```

Now the SA lookup keys off whatever SA the Deployment actually references; if that SA isn't the one carrying the annotations, `client_id` comes back empty and the existing assertion fails. I tested the shell/yq double-quote interpolation locally with mikefarah `yq` v4.53 — it resolves correctly.

The docs' federated-credential subject still has to spell the name out literally (it's an `az` CLI arg), so that duplication can't be removed — but treating `_helpers.tpl` as the one authoritative default and having CI derive from the render (rather than re-asserting a copy) is the meaningful part. A follow-up could add an explicit assertion that the Deployment's `serviceAccountName` equals the rendered ServiceAccount's name.

pkg/azurelustre/dynamic_provisioning.go
```diff
@@ -339,7 +342,15 @@ func (d *DynamicProvisioner) GetSkuValuesForLocation(ctx context.Context, locati
...
-			return nil, status.Errorf(codes.Internal, "error retrieving SKUs: %v", err)
+			st := status.Convert(convertHTTPResponseErrorToGrpcCodeError(err))
+			// Recognized HTTP errors keep their mapped code (e.g. 403 ->
+			// PermissionDenied with the auth hint). The helper maps unrecognized
+			// non-HTTP errors to Unknown, but this path reports Internal for those.
+			code := st.Code()
+			if code == codes.Unknown {
+				code = codes.Internal
+			}
+			return nil, status.Errorf(code, "error retrieving SKUs for location %s: %s", location, st.Message())
```
jeffbearer: **This `Unknown → Internal` remap diverges from the sibling ARM methods and quietly overrides a deliberate choice in the helper.** The helper's 5xx `default` case picks `codes.Unknown` on purpose — its own comment says *"Prefer to default to Unknown rather than Internal so provisioner will retry."* `CreateAmlFilesystem`/`DeleteAmlFilesystem` return `convertHTTPResponseErrorToGrpcCodeError(err)` verbatim and preserve that. Only this path rewrites `Unknown → Internal`, so an unmapped 5xx (or a non-`ResponseError`) during SKU listing surfaces as `Internal` while the exact same underlying error surfaces as `Unknown` from create/delete. Two error contracts for one concept.

If the intent is to align SKU listing with the other methods (and honor the helper's retry preference), just pass the mapped code through:

```suggestion
			st := status.Convert(convertHTTPResponseErrorToGrpcCodeError(err))
			return nil, status.Errorf(st.Code(), "error retrieving SKUs for location %s: %s", location, st.Message())
```

This keeps the win from this PR (403 → `PermissionDenied` with the auth hint still flows through) and drops only the special-casing. It does change the plain-error case (`errorLocation`) from `Internal` to `Unknown`, so the test at `dynamic_provisioning_test.go:1576` would need `expectedCode: codes.Unknown`.

Conversely, if preserving the legacy `Internal` for genuinely-unknown SKU errors is intentional (your comment on lines 346–348 suggests it is), that's a defensible call too — but then it's worth a one-line note on *why* SKU listing should differ from create/delete, so the next person doesn't "fix" the inconsistency by accident. Either way, the ask is just to make the divergence deliberate rather than incidental.

docs/workload-identity.md
```diff
@@ -0,0 +1,94 @@
...
+  --set IsWorkloadIdentityEnabled=Enabled \
+  --set IdentityClientId="${USER_ASSIGNED_CLIENT_ID}"
+# add --set IdentityTenantId="<TENANT_ID>" only for cross-tenant scenarios
+```
+
+### 6. Verify
+
+```bash
+kubectl logs -n kube-system deploy/csi-azurelustre-controller -c azurelustre \
+  | grep "authenticated with"
```
jeffbearer: Companion to my comment on `docs/csi-debug.md` (https://github.com/kubernetes-sigs/azurelustre-csi-driver/pull/311#discussion_r3659818940) — same release-name dependency, but in this file, so it needs its own fix.

`deploy/csi-azurelustre-controller` only resolves when the chart is installed with `fullnameOverride=csi-azurelustre` — the Deployment is otherwise named `{{ include "azurelustre.fullname" . }}-controller`. `csi-azurelustre-controller` is the stable *label*, not the object name. Use the release-agnostic selector (matching the pattern used throughout `csi-debug.md`):

```suggestion
kubectl logs -n kube-system -l app=csi-azurelustre-controller -c azurelustre --tail=300 \
  | grep -i "authenticated with"
```

Since this is the setup guide that points users at the troubleshooting docs, worth fixing both so they use the same robust form. (Step 6 also relies on the controller running at `-v=2`+ to surface the azidentity line, which it does by default via `controller.extraArgs: ["-v=5"]`, so the `grep` works out of the box.)

pkg/azurelustre/azurelustre.go
jeffbearer: Follow-up on test coverage for this gate: the flagship `TestNewDriverControllerIdentityModes` runs all three identity modes but asserts only `NoError`/`NotNil`, so it would stay green whether the predicate is `== ""` or `!= "all"`. And it can't be strengthened directly — `azcore/log` exposes only `SetEvents`/`SetListener` (no getter, no exported `Should`), so the global logging state this mutates isn't observable from a test.

The testable fix is to extract this decision into a pure predicate and unit-test that:

```go
// shouldForwardAzureAuthEvents reports whether the controller should forward
// azidentity Authentication events to klog. Defer to azcore's own logger when
// the operator enabled full SDK logging (AZURE_SDK_GO_LOGGING=all).
func shouldForwardAzureAuthEvents(sdkLogging string) bool {
	return sdkLogging != "all"
}
```

Call site becomes:

```go
if shouldForwardAzureAuthEvents(os.Getenv("AZURE_SDK_GO_LOGGING")) {
	azlog.SetEvents(azidentity.EventAuthentication)
	azlog.SetListener(func(event azlog.Event, msg string) {
		klog.V(2).Infof("azidentity %s: %s", event, msg)
	})
}
```

Test:

```go
func TestShouldForwardAzureAuthEvents(t *testing.T) {
	assert.True(t, shouldForwardAzureAuthEvents(""), "forward when SDK logging is unset")
	assert.True(t, shouldForwardAzureAuthEvents("true"), "forward for any value other than 'all'")
	assert.False(t, shouldForwardAzureAuthEvents("all"), "defer to azcore's own logger when full SDK logging is enabled")
}
```

This locks the decision from the comment above and would have caught the original gap: under `== ""`, `shouldForwardAzureAuthEvents("true")` returns `false`, so the `"true"` case fails. `TestNewDriverControllerIdentityModes` can stay as a construct-time smoke test.

dabradley: Removed the forwarding block entirely rather than fixing the gate. Two reasons: the SDK only emits the credential name on first token acquisition, not at startup, so there's no signal until something calls ARM; and the winning credential isn't queryable — ChainedTokenCredential.successfulCredential is unexported — so this could never be made reliable.

Instead, the chart now pins the credential per mode via AZURE_TOKEN_CREDENTIALS (WorkloadIdentityCredential when WI is enabled, ManagedIdentityCredential otherwise), so the identity is determined by configuration rather than discovered by a chain walk, and the driver logs it at startup with klog.Infof:

>controller authenticating with workload identity (client ID "…")

That also covers the verbosity comment: it's at default verbosity now, not V(2). shouldForwardAzureAuthEvents and its test went away with the block. This needed azidentity 1.9.0 → 1.14.0 for AZURE_TOKEN_CREDENTIALS support, which is upgraded in the previous commit.

docs/csi-debug.md
dabradley: Adopted both — the selector form for logs, and the POD=$(...) + exec pattern for the env check.

Related: since the SDK forwarding is gone (see the azurelustre.go thread), the message this step greps for no longer exists, so the command was dead as written. It now greps authenticating with and expects controller authenticating with workload identity (client ID "<client-id>"). Same fix in docs/workload-identity.md. Both commands were run verbatim against a live cluster in each identity mode to confirm they produce the documented output.

pkg/azurelustre/dynamic_provisioning.go
dabradley: Updated the three assertions you called out; also narrowed the SKU-listing test to match on "error retrieving SKUs"

hack/verify-helm-chart-files.sh
dabradley: Adopted. Worth checking that the default is what we expect and that it is used correctly in the template

pkg/azurelustre/dynamic_provisioning.go
dabradley: Implemented

docs/workload-identity.md
dabradley: Fixed the same way as csi-debug.md. The expected-output line also had to change: the SDK message it referenced is gone now that the credential is pinned, so step 6 greps authenticating with and expects controller authenticating with workload identity (client ID "<client-id>"). The "fell back to the node identity" note now points at a missing AZURE_FEDERATED_TOKEN_FILE injection, which is the actual cause.

charts/latest/azurelustre-csi-driver/templates/controller-deployment.yaml
dabradley: serviceAccount.create and the name overrides are now gone; the controller SA is always chart-managed with a fixed name, so this combination no longer exists.

The reasoning: the ServiceAccount name is the trust anchor for workload identity. It's embedded in the federated credential subject the operator configures out-of-band (system:serviceaccount:<ns>:csi-azurelustre-controller-sa), and the webhook keys injection off an annotation on that same SA. BYO-SA makes that annotation unverifiable by the chart; a configurable name makes the subject a moving target. Either way the chart can't guarantee the invariant workload identity depends on — which is exactly why your options reduced to hard-fail-a-valid-setup or silently-use-the-wrong-identity.

A static name removes the choice instead of picking a side, and follows the AKS extension guidance that service account names not be customer-configurable. Pre-release chart, so there's no migration cost.

If anyone needs to use this for that unsupported use-case, they're probably already making other changes to the configs/charts when using this repo.

charts/latest/azurelustre-csi-driver/templates/controller-deployment.yaml
dabradley: Implemented. Worth adding that the guard had zero test coverage as written, so hack/verify-helm-chart-files.sh now renders with IsWorkloadIdentityEnabled=enabled (lowercase) and asserts the render fails with the allowed-value message

docs/workload-identity.md
```diff
@@ -0,0 +1,91 @@
...
+### 6. Verify
+
+```bash
+kubectl logs -n kube-system -l app=csi-azurelustre-controller -c azurelustre --tail=300 \
+  | grep -i "authenticating with"
+# expect: controller authenticating with workload identity (client ID "<client-id>")
+```
+
+`authenticating with managed identity` means the webhook did not inject
+`AZURE_FEDERATED_TOKEN_FILE`, so the controller fell back to the node identity.
```
csmuell: I think this reads a little off.  Once workload identity is on, the controller is pinned to `WorkloadIdentityCredential` through `AZURE_TOKEN_CREDENTIALS` (see `controller-deployment.yaml:154`).

If the webhook never injects `AZURE_FEDERATED_TOKEN_FILE` it doesn't quietly drop back to the node identity -- the ARM calls just fail until the injection is sorted (by design).  This is really just a wording fix, nothing to change in the code.

maybe something like:

> `authenticating with managed identity` means the webhook didn't inject `AZURE_FEDERATED_TOKEN_FILE`. the controller's pinned to workload identity, so ARM calls will fail until that's fixed -- it won't fall back to the node identity.

same line in `docs/csi-debug.md:868`

csmuell (approved):
doc nit, lgtm.

docs/csi-debug.md
```diff
@@ -836,6 +836,45 @@ kubectl logs -n kube-system -l app=csi-azurelustre-controller -c azurelustre --t
...
+
+Confirm which identity the controller authenticates with:
+
+```sh
+kubectl logs -n kube-system -l app=csi-azurelustre-controller -c azurelustre --tail=300 | grep -i "authenticating with"
+# expect: controller authenticating with workload identity (client ID "<client-id>")
+```
+
+`authenticating with managed identity` means the webhook did not inject
+`AZURE_FEDERATED_TOKEN_FILE`, so the controller is using the node identity.
```
jeffbearer: Same wording issue @csmuell raised on `docs/workload-identity.md` (https://github.com/kubernetes-sigs/azurelustre-csi-driver/pull/311#discussion_r3685525054), but here in `csi-debug.md` — which that thread doesn't cover, so flagging it so both docs get corrected together.

With workload identity enabled the controller is pinned to `WorkloadIdentityCredential` via `AZURE_TOKEN_CREDENTIALS` (`controller-deployment.yaml:156-157`). I checked the vendored azidentity v1.14.0: when `AZURE_TOKEN_CREDENTIALS=WorkloadIdentityCredential`, the credential chain contains **only** workload identity — there is no `ManagedIdentityCredential` in it (`default_azure_credential.go`, `selected = workloadIdentity`). So a missing `AZURE_FEDERATED_TOKEN_FILE` makes token acquisition **fail**; the controller does **not** use the node identity. Suggested reword:

```suggestion
`authenticating with managed identity` means the webhook did not inject
`AZURE_FEDERATED_TOKEN_FILE`. Because the controller is pinned to
`WorkloadIdentityCredential` via `AZURE_TOKEN_CREDENTIALS` (no managed-identity
fallback), ARM calls then fail until injection is fixed — the controller does
not use the node identity.
```

Note the observable `authenticating with managed identity` log line itself comes from `azureIdentityConfigurationMessage()`, which infers the mode from the token file rather than the pinned credential — see my separate comment on `pkg/azurelustre/azurelustre.go` for the root cause.

pkg/azurelustre/azurelustre.go
```diff
@@ -271,6 +274,14 @@ func NewDriver(options *DriverOptions) (*Driver, error) {
 	return &d, nil
 }
 
+func azureIdentityConfigurationMessage() string {
+	clientID := os.Getenv("AZURE_CLIENT_ID")
+	if os.Getenv("AZURE_FEDERATED_TOKEN_FILE") != "" {
+		return fmt.Sprintf("controller authenticating with workload identity (client ID %q)", clientID)
+	}
+	return fmt.Sprintf("controller authenticating with managed identity (client ID %q)", clientID)
```
jeffbearer: Root cause of the doc-wording issues (see my comment on `docs/csi-debug.md` and @csmuell's on `docs/workload-identity.md`): this helper infers the identity mode from `AZURE_FEDERATED_TOKEN_FILE` presence, but post-pivot the active credential is pinned by `AZURE_TOKEN_CREDENTIALS` (`controller-deployment.yaml:156-157`). Those two disagree in exactly the failure case operators debug with this log line:

- WI enabled → `AZURE_TOKEN_CREDENTIALS=WorkloadIdentityCredential` pins the chain to WI only (verified in azidentity v1.14.0 — no managed-identity fallback).
- If the webhook doesn't inject `AZURE_FEDERATED_TOKEN_FILE`, this logs `authenticating with managed identity` — but the controller is **not** using managed identity; it's pinned to WI and will **fail** every ARM call.

So the one scenario where this message matters most, it reports the opposite of reality. Suggest keying off the pinned credential and calling out the missing-token failure explicitly:

```suggestion
	clientID := os.Getenv("AZURE_CLIENT_ID")
	// Report the credential the controller is actually pinned to via
	// AZURE_TOKEN_CREDENTIALS, not the token file's presence: when workload
	// identity is enabled the chain contains only WorkloadIdentityCredential
	// (no managed-identity fallback), so a missing token file is a failure, not
	// a switch to managed identity.
	if strings.EqualFold(os.Getenv("AZURE_TOKEN_CREDENTIALS"), "WorkloadIdentityCredential") {
		if os.Getenv("AZURE_FEDERATED_TOKEN_FILE") == "" {
			return fmt.Sprintf("controller pinned to workload identity (client ID %q) but AZURE_FEDERATED_TOKEN_FILE is not set; token acquisition will fail until the webhook injects it", clientID)
		}
		return fmt.Sprintf("controller authenticating with workload identity (client ID %q)", clientID)
	}
	return fmt.Sprintf("controller authenticating with managed identity (client ID %q)", clientID)
```

(`strings`, `fmt`, `os` are all already imported, so this compiles as-is. Optional: also handle an explicit `AZURE_TOKEN_CREDENTIALS=ManagedIdentityCredential` vs unset, but the above covers the misleading case.)

PR #312, opened 2026-08-05
feat: load Lustre client in a native sidecar; install utils in the dr…

dabradley:
…iver

The node DaemonSets previously ran one container that installed the full Lustre client and loaded the kernel modules before opening the CSI socket. On Azure Linux 3 that ~40s tdnf install exceeded the node-driver-registrar's hardcoded 30s connect deadline, so the registrar timed out and restarted on every node pod.

Split the work by container role (AZURELUSTRE_CSI_ROLE):

  - loader: a native sidecar (restartPolicy: Always) that installs the full metapackage, loads the kernel modules, configures LNet, then runs an LNet reconcile loop. Its startupProbe gates the driver and registrar so they never race a slow install, and a SIGTERM trap unloads the modules on teardown so nothing is left behind before a driver upgrade.
  - driver: installs only the kernel-agnostic userspace utils, then execs the CSI driver, relying on the loader having loaded the modules.

readinessProbe.sh becomes a loader-only LNet health check; the driver gates Ready on its CSI socket so a rolling upgrade cannot advance past a not-yet-serving node. AZURELUSTRE_CSI_ROLE (loader/driver/controller) is now required and replaces the AZURELUSTRE_CSI_INSTALL_LUSTRE_CLIENT gate; all three node flavors (azurelinux3, jammy, noble) share the pattern.

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


jeffbearer (commented):
One small, non-blocking suggestion below — the self-ping's internal `timeout` is looser than the probe budgets that invoke this script, so kubelet's timeout can pre-empt the script's own (more descriptive) failure message. Everything else in the PR looks solid to me; any other notes I have are discussion-level rather than blocking.

pkg/azurelustreplugin/readinessProbe.sh
```diff
@@ -14,72 +14,56 @@
...
+first_nid=$(lnetctl net show 2>/dev/null | grep "nid:" | grep -v "@lo" | sed -n '1s/.*nid: \([^ ]*\).*/\1/p' || echo "")
+if [[ -z "${first_nid}" ]]; then
     echo "Unable to determine LNet NID for self-ping test"
     exit 1
 fi
-
-# Perform self-ping test with timeout
-if ! timeout 10 lnetctl ping "${FIRST_NID}" >/dev/null 2>&1; then
-    echo "LNet self-ping test failed for NID: ${FIRST_NID}"
+if ! timeout 10 lnetctl ping "${first_nid}" >/dev/null 2>&1; then
```
jeffbearer: **Bound the self-ping below the probe timeout so the script's own error wins.**

This `timeout 10` is looser than every probe that runs this script: the loader `startupProbe` has `timeoutSeconds: 5` (so kubelet SIGKILLs the exec at 5s before this fires), and the `readinessProbe`'s `timeoutSeconds: 10` ties the internal `timeout`, racing it. In both cases an opaque kubelet timeout can pre-empt the descriptive `LNet self-ping test failed for NID: …` message below. Dropping the guard strictly under the tightest probe budget (5s) makes this script authoritative and keeps the clean error — a local self-ping normally returns in well under a second.

```suggestion
# Keep this timeout strictly below the tightest probe timeoutSeconds (the loader
# startupProbe's 5s) so this script's own guard fires first and emits the
# descriptive message below, instead of kubelet SIGKILLing the exec on timeout.
# The whole script shares that budget; a local self-ping normally returns in
# well under a second.
if ! timeout 3 lnetctl ping "${first_nid}" >/dev/null 2>&1; then
```

jeffbearer (commented):
**Optional / non-blocking — registrar startup hardening (feel free to punt).**

Not a blocker, and the current config already has adequate margin — flagging this only as a cheap belt-and-suspenders for later.

The `node-driver-registrar` restart isn't a connect timeout (`csi-lib-utils` `Connect()` retries the socket indefinitely); it's driven by its registration `livenessProbe` (`--mode=kubelet-registration-probe`, `initialDelaySeconds: 30`, plus the defaults `periodSeconds: 10` / `failureThreshold: 3` → effective restart deadline ~50–60s). With the loader/driver split the driver opens its socket in ~18s and registers comfortably inside that window, so this is fine today.

The only latent risk is that the window is **implicit**: if the driver's utils-install time grows (slower mirror, larger package), registration could creep toward the liveness deadline and restart the registrar on startup.

If you'd like to make that window explicit and generous, a `startupProbe` on the registrar (same registration probe) gates the existing `livenessProbe`, so install-time variance can't trip a startup restart:

```yaml
          startupProbe:
            exec:
              command:
                - /csi-node-driver-registrar
                - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
                - --mode=kubelet-registration-probe
            initialDelaySeconds: 5
            periodSeconds: 5
            timeoutSeconds: 15
            failureThreshold: 30   # ~150s to register before liveness engages
```

It would go in the `node-driver-registrar` container of the three `deploy/csi-azurelustre-node-*.yaml` manifests and the three `charts/latest/azurelustre-csi-driver/templates/node-daemonset-*.yaml` templates (the registrar block is identical across flavors); the existing `livenessProbe` stays for steady state. I couldn't attach this as a one-click suggestion because this PR doesn't otherwise touch the registrar block — hence the paste-in snippet. Purely an optional pickup.

pkg/azurelustreplugin/readinessProbe.sh
dabradley: This is reasonable, I'll update that

dabradley:
> **Optional / non-blocking — registrar startup hardening (feel free to punt).**
> 
> Not a blocker, and the current config already has adequate margin — flagging this only as a cheap belt-and-suspenders for later.
> 
> The `node-driver-registrar` restart isn't a connect timeout (`csi-lib-utils` `Connect()` retries the socket indefinitely); it's driven by its registration `livenessProbe` (`--mode=kubelet-registration-probe`, `initialDelaySeconds: 30`, plus the defaults `periodSeconds: 10` / `failureThreshold: 3` → effective restart deadline ~50–60s). With the loader/driver split the driver opens its socket in ~18s and registers comfortably inside that window, so this is fine today.
> 
> The only latent risk is that the window is **implicit**: if the driver's utils-install time grows (slower mirror, larger package), registration could creep toward the liveness deadline and restart the registrar on startup.
> 
> If you'd like to make that window explicit and generous, a `startupProbe` on the registrar (same registration probe) gates the existing `livenessProbe`, so install-time variance can't trip a startup restart:
> 
> ```yaml
>           startupProbe:
>             exec:
>               command:
>                 - /csi-node-driver-registrar
>                 - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
>                 - --mode=kubelet-registration-probe
>             initialDelaySeconds: 5
>             periodSeconds: 5
>             timeoutSeconds: 15
>             failureThreshold: 30   # ~150s to register before liveness engages
> ```
> 
> It would go in the `node-driver-registrar` container of the three `deploy/csi-azurelustre-node-*.yaml` manifests and the three `charts/latest/azurelustre-csi-driver/templates/node-daemonset-*.yaml` templates (the registrar block is identical across flavors); the existing `livenessProbe` stays for steady state. I couldn't attach this as a one-click suggestion because this PR doesn't otherwise touch the registrar block — hence the paste-in snippet. Purely an optional pickup.

The timeouts / checks with the registrar are confusing. The main reason for this change is because trying to do that doesn't actually prevent the registrar from reaching its internal non-overridable 30 second timeout. The issue is that when our driver takes a long time to be ready to serve requests (> 30s), it's not these probes that cause it to restart. Usually, the registrar probes are basically a no-op since it's able to almost immediately attempt to communicate on the registration path. So, its probes are fine because it's able to start probing for our driver almost immediately. Setting longer probes won't prevent any behavior, since they aren't actually failing if our driver isn't available yet. Since the registrar can almost immediately start communicating on the registration path, it starts its own internal 30 second timeout for our driver to respond. When our driver doesn't respond in that much time, it exits its own process, (causing a restart).

Note how the registrar works https://github.com/kubernetes-csi/node-driver-registrar/blob/v2.13.0/cmd/csi-node-driver-registrar/main.go#L174

```go
	csiConn, err := connection.ConnectWithoutMetrics(ctx, *csiAddress)
	if err != nil {
		logger.Error(err, "Error connecting to CSI driver")
		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
	}
```

Which defaults to 30 seconds:
```go
func ConnectWithoutMetrics(ctx context.Context, address string, options ...Option) (*grpc.ClientConn, error) {
	// Prepend default options
	options = append([]Option{WithTimeout(time.Second * 30)}, options...)
	return connect(ctx, address, options)
}
```

With the split as it is, doing the installation and setup in the loader offloads enough of the setup cost to the initialization phase that the driver's work can fall within the 30 second window. If we find that we're still doing too much in the driver, we would have to solve that by either baking the lustre utils into the image somehow or having a shared volume that can be set up via an init container or something. The current approach appears sufficient for the goal of having the drivers able to start quickly enough though, as you mention.

In the case that the package repository is unusually slow, or the package otherwise takes too long to install, yes, we can have registrar restarts, but they will not prevent the successful initialization of the driver. The driver healthcheck is still independent of what the registrar is doing. I am considering if it's worth still including a startupProbe for the driver, just to handle a slow package repo case independently of the ongoing livenessProbe once it's gotten everything it needs.

This question did make me realize that the registrar's actual healthcheck probe has changed since we first configured it. The old way (that we're using) is a no-op and always succeeds now. I'll make a note to update this so it's more accurate, but as I mentioned above, none of his affects how it will react to a slower driver container start.

pkg/azurelustreplugin/readinessProbe.sh
dabradley: New push addresses this

PR #315, opened 2026-08-17
feat: keep CSI node plugin off reserved system pools and comply with AKS Automatic admission policy

jeffbearer:

Makes the Azure Managed Lustre CSI driver schedule correctly on AKS — including **AKS Automatic** — as it drives toward GA.

The node plugin does privileged host/kernel work (loads the Lustre client module, configures host LNet, bind-mounts into the kubelet path), so it must run on the **user node pools** where Lustre workloads land and stay off AKS-managed **reserved/tainted** system pools — without silently disappearing on a cluster whose only pool is an *untainted* system pool.

This uses the AKS-recommended pattern for keeping a DaemonSet off system pools ([aka.ms/aks/use-system-pools](https://aka.ms/aks/use-system-pools)): **do not tolerate the `CriticalAddonsOnly` taint** rather than hard-excluding system nodes by label.

> **Note:** This replaces the earlier affinity-based approach in this PR. An adversarial review found that a required `mode NotIn [system]` affinity zeroed out the node plugin on a stock single-system-pool cluster (silent mount failure), and a blanket `{operator: Exists}` toleration is rejected on AKS Automatic. This revision fixes both.

## Validation / testing

- `hack/verify-helm-chart-files.sh` → **pass** (chart ↔ `deploy/*.yaml` parity maintained)
- `helm lint` → **0 failed**
- Rendered-manifest assertions: **0** `CriticalAddonsOnly` tolerations, **0** `kubernetes.azure.com/mode` affinity, spot + `nvidia.com/gpu` tolerations present on all 3 node DaemonSet flavors
- **Live AKS Automatic** (k8s 1.35.6, MLT-10 smoke sub): the **final keyed-toleration chart** (this PR's exact commits) was `helm install`ed into the safeguards-excluded `azurelustre-system` namespace as a non-exempt Entra user identity. Results:
  - **Admission**: the node DaemonSets and controller Deployment were **admitted past the always-on `aks-managed-critical-addons-only` VAP** (server dry-run + real install) — the same policy that *rejected* the earlier blanket-`Exists` / `CriticalAddonsOnly` design.
  - **Node plugin**: landed only on a NAP-provisioned untainted `mode=user` node (Azure Linux 3 flavor); **0 pods on any of the 6 `mode=system` nodes** and off the not-ready node.
  - **Controller**: first reported `FailedScheduling — 6 node(s) had untolerated taint(s)` (correctly refusing every system node now that `CriticalAddonsOnly` is dropped), then scheduled onto the untainted `mode=user` node per the flipped `mode NotIn [system]` soft preference.
  - Test cluster torn down after validation.

**Known limitation:** the live run above exercised an earlier build of this branch (empty node tolerations) to prove the admission + namespace-exclusion path. The final **keyed** toleration set here is verified statically (lint + parity + render) — keyed `Equal`/`Exists` tolerations are compliant because the policy only bans `CriticalAddonsOnly`, keyless `{operator: Exists}`, and `kubernetes.azure.com/hostedvm`.

## Changes

**Node DaemonSets** (`jammy` / `noble` / `azurelinux3`, chart + `deploy/`):

- Replace the blanket `{operator: Exists}` toleration with **keyed tolerations** for the common AKS user-pool taints (spot: `kubernetes.azure.com/scalesetpriority=spot`, GPU: `sku=gpu` and `nvidia.com/gpu`).
- Not tolerating `CriticalAddonsOnly` keeps the plugin off reserved/tainted system pools (incl. AKS Automatic) while still running on an untainted single-system-pool cluster.

**Controller Deployment** (chart + `deploy/`):

- Remove the `CriticalAddonsOnly` toleration and **flip** the soft node-affinity preference from `kubernetes.azure.com/mode In [system]` to `NotIn [system]`, so the controller now *prefers user pools* and is steered off AKS-managed system pools when user capacity exists.
- This soft (`preferredDuringScheduling`) preference is meaningful for the controller because it is a **Deployment** — the scheduler can place its replicas. (A `preferred` node affinity would be a no-op on the node **DaemonSet**, which is why the node plugin is taint-based instead.)

**Documentation:**

- `docs/install-csi-driver.md`: new "Node pool taints and tolerations" section, with an IMPORTANT callout that custom-tainted Lustre pools need an added `node.tolerations` entry (else silent mount failures).
- `charts/README.md`: add `node.tolerations` / `controller.tolerations` rows to the parameters table.

Pod Security / Deployment Safeguards compliance is unchanged — it is satisfied by running in a safeguards-excluded namespace and is independent of tolerations and node placement (no `securityContext` changes here).

## Accepted trade-offs

- Customers who taint a Lustre user pool with something **other** than spot/GPU must add a matching toleration via `node.tolerations` (documented in `values.yaml`). This is the deliberate cost of admission compliance vs. the rejected blanket `Exists`.
- On an **untainted** AKS Standard system pool, the node plugin will still run there (by design — the goal is avoiding *reserved/tainted* nodes).

## Does this PR introduce a user-facing change?

```release-note
Azure Lustre CSI driver: the node plugin now uses keyed tolerations for common AKS user-pool taints (spot, GPU) instead of a blanket toleration, and no longer tolerates the CriticalAddonsOnly taint, so it is admitted on AKS Automatic and stays off reserved system node pools. The controller no longer targets AKS-managed system nodes. Customers using custom user-pool taints should set node.tolerations.
```


dabradley (approved):
/lgtm

PR #317, opened 2026-08-20
Update release doc

dabradley:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


charts/README.md
```diff
@@ -1,20 +1,24 @@
 # Install Azure Lustre CSI Driver with Helm 3
 
-## Add Helm Repo
+## Released chart versions
 
-To add the Helm repo:
+| Chart version | Driver image family |
+| --- | --- |
```
jeffbearer, line 6: Optional discovery aid (non-blocking): a way to see what's actually published, while keeping this table the source of truth for supported versions.

```suggestion
| --- | --- |

Install only versions listed above. To see every chart tag currently published to the
registry — including pre-release and customer-hidden preview tags that are **not**
supported for general install — list them directly:

    curl -s https://mcr.microsoft.com/v2/microsoft.azuremanagedlustre/azurelustre-csi-driver/tags/list | jq -r '.tags[]'
```

RELEASE.md
```diff
@@ -6,103 +6,513 @@ Container Upstream build system), which provides FIPS compliance, supply-chain
...
+| `rolloutType` | `normal` |
+
+The pipeline must complete these gates in order:
+
+1. Clone the chart from the reviewed CSI commit.
+2. Set chart `version` to `A.B.C`; independently set `appVersion` and
+  `image.tag` to driver image family `vX.Y.Z`.
+3. Render the chart and verify every workload image resolves from MCR.
+4. Publish the immutable OCI chart to
+  `mcr.microsoft.com/microsoft.azuremanagedlustre/azurelustre-csi-driver:A.B.C`.
```
jeffbearer: Small precision suggestion to keep this in lockstep with the publisher/operator guide's delivery model. The pipeline pushes the chart to the **team ACR**; the **MCROnboard webhook** (asynchronous, out-of-band) is what copies it to MCR — which is exactly why gate 5 then *waits* for the public MCR digest to match the ACR digest. Wording gate 4 as a direct pipeline→MCR publish slightly obscures that the syndication is a separate async step (and that the chart lands on MCR via the webhook even under `publishOnly: true`).

```suggestion
4. Publish the immutable OCI chart to the team ACR; the MCROnboard webhook then
  syndicates it to
  `mcr.microsoft.com/microsoft.azuremanagedlustre/azurelustre-csi-driver:A.B.C`.
```

charts/README.md
dabradley: Applied this wording

RELEASE.md
dabradley: Applied this wording

PR #322, opened 2026-08-28
feat: support optional digest pinning for CSI sidecar images

jeffbearer:
**What this PR does / why we need it**

Adds an optional per-sidecar image digest to the chart so deployed workloads can pull the CSI sidecar images (`csi-provisioner`, `livenessprobe`, `csi-node-driver-registrar`) by immutable `sha256` digest instead of by mutable tag, while keeping the human-readable tag in the reference. This is supply-chain hardening for first-party (AKS extension) publishing; it is fully opt-in and default-off.

A new `azurelustre.imageRef` helper renders `<repo>:<tag><suffix>` and, when a digest is supplied, trims it, validates it as `sha256:<64 hex>`, and appends `@<digest>`. A malformed digest fails the render locally instead of surfacing later as a runtime `ImagePullBackOff`. `values.yaml` gains an empty `digest` field on each sidecar.

**Why the driver image is intentionally NOT digest-pinnable**

The driver image is published by the first-party DALEC pipeline, which republishes CVE and dependency fixes under the same release tag (e.g. `v0.5.0-noble` -> `-1`, `-2`, ...; the prior release accumulated 30+ such rebuilds). Pinning the driver to a digest would freeze it on a fixed base image and suppress those automatic CVE refreshes until a full patch release + customer upgrade — undermining DALEC's model. Driver supply-chain integrity is instead provided by DALEC signing/provenance/SBOM. Sidecar version tags are immutable (no such republish), so pinning them is pure upside.

**Special notes for your reviewer**

Empty digests reproduce the exact previous tag-only references, so rendered output is byte-for-byte unchanged by default and `deploy/` manifests stay in sync (verified via `hack/verify-helm-chart-files.sh` and `make verify`).

**Does this PR introduce a user-facing change?**

```release-note
The Helm chart now supports optional immutable digest pinning for the CSI sidecar images via `sidecars.<name>.digest`. Default behavior is unchanged (images are still referenced by tag when no digest is set).
```

### DIFF
--- a/charts/latest/azurelustre-csi-driver/templates/_helpers.tpl
+++ b/charts/latest/azurelustre-csi-driver/templates/_helpers.tpl
@@ -66,3 +66,25 @@ imagePullSecrets:
   {{- end }}
 {{- end }}
 {{- end -}}
+
+{{/*
+Render a container image reference.
+Args (dict): repository (required), tag (required), suffix (optional flavor
+suffix, e.g. "-noble"), digest (optional "sha256:...").
+Renders "<repository>:<tag><suffix>". When digest is set, it is trimmed and
+validated as "sha256:<64 hex>" then appended as "@<digest>" so the image is
+pulled by immutable digest while the tag stays human-readable. A malformed
+digest fails the render (fast, local) instead of surfacing as a runtime
+ImagePullBackOff. Empty digest reproduces the historical tag-only reference.
+*/}}
+{{- define "azurelustre.imageRef" -}}
+{{- $ref := printf "%s:%s%s" .repository .tag (default "" .suffix) -}}
+{{- with .digest -}}
+{{- $d := trim . -}}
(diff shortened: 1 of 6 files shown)


jeffbearer:
Closing — reversing course on digest pinning.

The DALEC first-party pipeline that publishes these images republishes CVE, OS, and dependency fixes under the same (floating) release tags for **all** images it builds — the CSI sidecars included, not just the driver. Pinning any image to a digest would freeze it on a fixed base and suppress those automatic security refreshes until a full patch release, undermining DALEC's entire security-patching model.

An earlier revision of this PR pinned sidecars on the assumption their tags were immutable; that assumption is incorrect for DALEC-built images. Since neither the driver nor the sidecars should be digest-pinned, this change is withdrawn and the chart stays tag-only. Rationale is captured in the extension security assessment (risk-accepted with MCR content-trust + DALEC continuous rebuild as compensating controls).

PR #327, opened 2026-09-10
Fix/azure cloud support

romahamu:
**What type of PR is this?**
Azure sovereign-cloud dynamic provisioning fix

/kind bug
This PR fixes AMLFS CSI dynamic provisioning for AKS clusters running in Azure China.

• Uses the Azure cloud configured in  azure.json  when creating Azure credentials and ARM clients, so Azure China uses the correct ARM endpoint and token audience instead of the public-cloud endpoint.
• Adds the controller-only  --allow-unadvertised-zones  flag, disabled by default.
• Allows an explicitly supplied availability zone to be sent to the AMLFS create API when the SKU API incorrectly returns no zones.
• Preserves normal zone validation when the SKU API returns a non-empty zone list.
• Adds unit tests and documentation for Helm and direct manifest installations.
• Validated dynamic provisioning in  chinanorth3  with  AMLFS-Durable-Premium-125  and zone  "3" ; the PVC bound and AMLFS reached  Succeeded  /  Available .

Fixes #39669561

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/docs/dynamic-provisioning.md
+++ b/docs/dynamic-provisioning.md
@@ -14,12 +14,33 @@ kubelet identity must have the correct permissions granted to it.
 Ensure that the kubelet identity has all of the permissions that are listed in the section
 on [Permissions For Kubelet Identity](driver-parameters.md#permissions-for-kubelet-identity).
 
+### Azure cloud environment
+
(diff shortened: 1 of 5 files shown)


dabradley (approved):
/lgtm

